PR #1, opened 2019-10-25
New test: create data sample located in server path

samlesu:
We can add data sample using two different modes:
- data file located on the user machine
- data file located on the server machine

The latest case was not tested before... :)


tests/conftest.py
```diff
@@ -102,6 +102,13 @@ def data_network():
         return f, n
 
 
+@pytest.fixture
+def node_cfg_1():
+    """Node configuration (first node)."""
+    cfg = settings.load()
+    return cfg.nodes[0]
+
+
```
jmorel: This fixture is the same as `node_cfg` but isn't used anywhere (yet). We should probably remove it to avoid confusion.

tests/conftest.py
```diff
@@ -114,6 +121,13 @@ def session_2(network):
     return network.sessions[1]
 
 
+@pytest.fixture
+def node_cfg(network):
```
jmorel: The network arg isn't used.

tests/values-docker-compose.yaml
```diff
@@ -4,8 +4,10 @@ nodes:
     address: 'http://owkin.substra-backend:8000'
     user: 'admin'
     password: 'admin'
+    shared_path: '/tmp/substra/'
```
jmorel: The new test doesn't pass when using docker-compose, I get the following error:
```
E               substra.sdk.exceptions.InvalidRequest: 400 Client Error: Bad Request for url: http://owkin.substra-backend:8000/data_sample/: One of your paths does not exist, is not a directory or is not an absolute path: /tmp/substra//gngnvl7v/data-0
```

I guess this is because of the trailing slash on this line.

samlesu: Yes it has not been tested with docker-compose, I'm not sure where medias are located with the latest docker-compose and the latest mac update.

jmorel: You could use the `SUBSTRA_PATH` env variable for that with a default on `/substra`. This is what's used by the `start.py` script in the backend, so it should match.

samlesu: I think the tests setup should stay independent from the backend setup, so we should update directly the `values-docker-compose.yaml` file.

If really needed we could use a test env var for that but it should be named accordingly (`SUBSTRAT_SHARED_PATH`) .

tests/settings.py
```diff
@@ -52,14 +53,15 @@ def _load_skaffold(path):
...
-    backends = [s for s in services if s['name'].startswith('substrabac-peer')]
+    backends = [s for s in services if s['name'].startswith('substra-backend-peer')]
 
     nodes = [NodeCfg(
         name=b['name'],
         msp_id=b['overrides']['peer']['mspID'],
         address=b['overrides']['backend']['defaultDomain'],
         user=b['overrides']['backend']['auth']['user'],
         password=b['overrides']['backend']['auth']['password'],
+        shared_path=b['overrides']['persistence']['hostPath'],
```
AurelienGasser: @samlesu from [your comment](https://github.com/SubstraFoundation/substra-tests/pull/1/files#r339546891):

> I think the tests setup should stay independent from the backend setup

Following the same logic, would it make sense to hardcode the tmp path in the skaffold case, instead of pulling it from the skaffold file? 

i.e. 

```suggestion
        shared_path="/tmp/substra",
```

samlesu: All the network configuration is created from the skaffold file directly so I don't think it is a good idea to hardcode only the `shared_path`.

In my opinion, what could be done for docker-compose is to add a parser for a `docker-compose.yaml` file generating the test configuration.

AurelienGasser: If I understand correctly, there would be 3 ways to launch the tests:

- with a `docker-compose.yaml` file
- with a `skaffold.yaml` file
- with a custom `config.yaml` file

Is that correct?

samlesu: My previous comment may not be very clear. Let me add a bit of context :)

Currently there are two different ways to specify the network configuration:
- create a values file and define explicitly all the values for each node (works for both docker-compose and skaffold)
- parse the backend skaffold file to generate the network configuration

From my point of view, we need to keep this two modes and the first mode should be independent from how the network is started.

samlesu: > If I understand correctly, there would be 3 ways to launch the tests:
> 
> * with a `docker-compose.yaml` file
> * with a `skaffold.yaml` file
> * with a custom `config.yaml` file
> 
> Is that correct?

Exactly.

tests/conftest.py
samlesu: Done

tests/conftest.py
samlesu: Done

tests/values-docker-compose.yaml
jmorel: If we want to keep things simple, we could just remove the trailing slash so that the generated paths are `/tmp/substra/gngnvl7v/data-0` instead of `/tmp/substra//gngnvl7v/data-0`.

AurelienGasser: What about this (until we have a better solution)

```diff
-    shared_path: '/tmp/substra/'
+   shared_path: '/tmp/substra' # needs to match SUBSTRA_PATH
```

samlesu: Looks like a good solution, will do that.

tests/test_network.py
```diff
@@ -39,6 +43,19 @@ def test_add_data_sample(factory, session):
...
 
 
+def test_add_data_sample_located_in_shared_path(factory, session, node_cfg):
+    spec = factory.create_dataset()
+    dataset = session.add_dataset(spec)
+
+    spec = factory.create_data_sample(test_only=True, datasets=[dataset])
+
+    # move data sample to substra backend shared file system
+    with tempfile.TemporaryDirectory(prefix=node_cfg.shared_path + '/') as dst:
```
samlesu: We should not delete data sample path.
When adding assets in other tests their data are not deleted, it should be the same here.

If needed deleting the node "data" files could be done at the end of each tests, all tests? But this should be done in a different PR.

tests/values-docker-compose.yaml
jmorel: 👍

tests/test_network.py
```diff
@@ -39,6 +42,22 @@ def test_add_data_sample(factory, session):
...
+def test_add_data_sample_located_in_shared_path(factory, session, node_cfg):
+    spec = factory.create_dataset()
+    dataset = session.add_dataset(spec)
+
+    spec = factory.create_data_sample(test_only=True, datasets=[dataset])
+
+    # move data sample to substra backend shared file system
+    dst = node_cfg.shared_path
+    dst = dst if dst.endswith('/') else dst + '/'
+    shutil.move(spec.path, dst)
```
jmorel: While running the tests with docker compose, this line fails with the following error:
`E               shutil.Error: Destination path '/tmp/substra/data-0' already exists`

samlesu: Good catch, the tests suite can run only once :)

samlesu: Done

samlesu: Data samples are now moved to a unique directory:
```
/tmp/peer-1
├── 2r4bixf2
│   └── data-0
│       └── data.csv
├── 63s3g6ab
│   └── data-0
│       └── data.csv
├── ax5vulhq
│   └── data-0
│       └── data.csv
```

AurelienGasser:
The tests fails on my machine with:

```
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/peer-1/ehvwo9vt'
```

See detailed log [here](https://gist.github.com/AurelienGasser/30e0ec78c98dedb4c68627adec70b0b9).

I suspect something is misconfigured on my machine (even though I have pulled the `dev` branches everywhere).

GuillaumeCisco:
I had the same issue.
I rerun the tests and got an error on, guess what `/tmp/peer-2`.
I rerun a new time and it passed.

samlesu:
Thanks for the feedback! That's weird, I cannot replicate on my machine, coud be related to the OS?

I used the option `prefix` instead of  `dir`  when calling `tempfile.mkdtemp()` (https://docs.python.org/3/library/tempfile.html#tempfile.mkdtemp). Maybe it could fix this issue?

AurelienGasser:
@samlesu I tested again after manually creating the `/tmp/peer-1` and `/tmp/peer-2` folders.

I'm getting a new error:

```
400 Client Error: Bad Request for url: http://substra-backend.node-1/data_sample/
```

```
substra.sdk.rest_client:rest_client.py:80 Requests error status 400: {"message":"One of your paths does not exist, is not a directory or is not an absolute path: /tmp/peer-1/tmpumecz1ln/data-0"}
```

samlesu:
@ClementGautier it looks like that the backend persistentVolumeClaim for the medias folder (`/tmp/peer-X`) is not mounted in all setup. It's correctly mounted on my setup (docker-for-mac and its integrated kubernetes cluster) but not for Aurélien and Guillaume.

I now remember that a workaround to avoid using the shared medias folder has been implemented in the populate for skaffold.

Do you have an idea what could cause the issue?

ClementGautier:
@samlesu Yes, you need to activate the `storage-provisioner` of minikube in order for the volume to be created. Also note that those directory are cleaned by a job launched during the helm delete but if you kill the job before it is able to do its job you might encounter some issues in your following attempts

jmorel:
@samlesu There are a couple conflicts that you need to revolve first.

PR #8, opened 2019-10-31
Add composite algo and traintuple to tests helper

samlesu:
Associated PRs:
- [cli](https://github.com/SubstraFoundation/substra/pull/5)
- [backend](https://github.com/SubstraFoundation/substra-backend/pull/13)
- [chaincode](https://github.com/SubstraFoundation/substra-chaincode/pull/5)
- [tools](https://github.com/SubstraFoundation/substra-tools/pull/3)

### DIFF
--- a/substratest/assets.py
+++ b/substratest/assets.py
@@ -8,6 +8,9 @@ import typing
 from . import errors
 
 
+FUTURE_TIMEOUT = 120  # seconds
+
+
 class Future:
     """Future asset."""
     # mapper from asset class name to client getter method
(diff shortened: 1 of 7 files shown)


substratest/factory.py
```diff
@@ -157,6 +185,20 @@ class TraintupleSpec(_Spec):
...
+    data_manager_key: str
+    train_data_sample_keys: str
+    in_head_model_key: str
+    in_trunk_model_key: str
+    tag: str
+    compute_plan_id: str
+    out_trunk_model_permissions: typing.Dict
+    rank: int = None
+
+
```
Kelvin-M: It could be nice to have a  `_TraintupleSpec` as you have done for `_Algo`

substratest/assets.py
```diff
@@ -233,6 +244,33 @@ class Meta:
...
+    rank: int
+    tag: str
+    log: str
+    in_head_model: InModel = None
+    in_trunk_model: InModel = None
+    out_head_model: OutCompositeModel = None
+    out_trunk_model: OutCompositeModel = None
+
+    class Meta:
+        mapper = {
```
Kelvin-M: It could be nice to have a  `_Traintuple` as you have done for `_Algo`

substratest/factory.py
samlesu: Yes I thought about this and I'm not 100% sure about what we should do.
For the `Algo` it makes more sense as all the fields are common for an `Algo` and a `CompositeAlgo`.
Thi is not true for `Traintuple` / `CompositeTraintuple`. I would wait a refactoring and finishing the implementation of the `AggregateTraintuple` before factorizing code.

jmorel:
@samlesu I pushed a fix removing a testtuple on an aggregatuple because it's not supported.

PR #16, opened 2019-11-07
add list tests

samlesu:


substratest/assets.py
```diff
@@ -120,16 +121,24 @@ class Permissions(_DataclassLoader):
...
     class Meta:
         mapper = {
             'pkhash': 'key',
         }
 
 
+@dataclasses.dataclass(frozen=True)
+class DataSample(_Asset):
+    key: str
+    owner: bool
```
inalgnu: isn't str expected here?

samlesu: good catch

PR #19, opened 2019-11-18
Add aggregate algo/traintuple to tests helper

samlesu:
Missing test using aggregate algo / traintuple

### DIFF
--- a/substratest/factory.py
+++ b/substratest/factory.py
@@ -65,6 +65,22 @@ if __name__ == '__main__':
     tools.algo.execute(TestAlgo())
 """
 
+DEFAULT_AGGREGATE_ALGO_SCRIPT = f"""
+import json
+import substratools as tools
+class TestAggregateAlgo(tools.AggregateAlgo):
+    def aggregate(self, models, rank):
+        return [0, 66]
+    def load_model(self, path):
+        with open(path) as f:
+            return json.load(f)
+    def save_model(self, model, path):
+        with open(path, 'w') as f:
+            return json.dump(model, f)
+if __name__ == '__main__':
+    tools.algo.execute(TestAggregateAlgo())
+"""
+
 DEFAULT_COMPOSITE_ALGO_SCRIPT = f"""
 import json
 import substratools as tools
@@ -173,6 +189,11 @@ class AlgoSpec(_AlgoSpec):
 
 
 @dataclasses.dataclass
+class AggregateAlgoSpec(_AlgoSpec):
+    pass
+
+
+@dataclasses.dataclass
 class CompositeAlgoSpec(_AlgoSpec):
     pass
 
@@ -183,6 +204,17 @@ class TraintupleSpec(_Spec):
     objective_key: str
     data_manager_key: str
     train_data_sample_keys: str
+    in_models_keys: str
+    tag: str
+    compute_plan_id: str
+    rank: int = None
+
+
+@dataclasses.dataclass
+class AggregateTraintupleSpec(_Spec):
+    algo_key: str
+    objective_key: str
+    worker: str
     in_models_keys: str
     tag: str
     compute_plan_id: str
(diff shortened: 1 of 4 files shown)


substratest/factory.py
```diff
@@ -415,6 +453,24 @@ def create_traintuple(self, algo=None, objective=None, dataset=None,
             rank=rank,
         )
 
+    def create_aggregate_traintuple(self, algo=None, objective=None, worker=None,
+                                    traintuples=None, tag=None,
+                                    compute_plan_id=None, rank=None):
+        traintuples = traintuples or []
+
+        for t in traintuples:
+            assert isinstance(t, assets.Traintuple)
```
samlesu: How do we support composite traintuple? 
Only trunk model can be aggregated.

tests/test_execution.py
```diff
@@ -245,3 +245,124 @@ def test_composite_traintuples_execution(factory, session):
...
+    traintuples = []
+    for data_sample in train_data_samples:
+        spec = factory.create_traintuple(
+            algo=algo,
+            objective=objective,
+            dataset=dataset,
+            data_samples=[data_sample],
+        )
+        traintuple = session.add_traintuple(spec).future().wait()
+        traintuples.add(traintuple)
```
jmorel: I guess it's `traintuples.append`.

tests/test_execution.py
```diff
@@ -245,3 +245,124 @@ def test_composite_traintuples_execution(factory, session):
...
+
+
+def test_aggregate_traintuples(factory, session):
+    """Aggregate traintuples."""
+    number_of_traintuples_to_aggregate = 3
+    spec = factory.create_dataset()
+    dataset = session.add_dataset(spec)
+
+    train_data_samples = []
+    for i in range(number_of_traintuples_to_aggregate):
```
jmorel: My linter lights up because of this unused `i`. Could we use `_` instead?

tests/test_execution.py
```diff
@@ -245,3 +245,124 @@ def test_composite_traintuples_execution(factory, session):
...
+        in_models_keys=traintuples,
+    )
+    aggregatetuple = session.add_aggregatetuple(spec).future().wait()
+    assert aggregatetuple.status == 'done'
+
+
+def test_aggregate_composite_traintuples(factory, session_1, session_2):
+    """Do 2 rounds of composite traintuples aggregations on multiple nodes.
+
+    This test refers to the model composition use case.
```
jmorel: Would there be a way to include a quick description of the use case? Just to make clear what we're executing here. Maybe some asciiart? I'm half joking about this, since we're creating tuple-graphs, it'd be very nice to have a way to represent these graphs inside the code.

tests/test_execution.py
samlesu: Done

tests/test_execution.py
samlesu: Done

tests/test_execution.py
samlesu: I fully agree, this is quite hard to understand what is the goal of this test.

I added a detailed description of the compute plan.

tests/test_execution.py
```diff
@@ -245,3 +245,142 @@ def test_composite_traintuples_execution(factory, session):
...
+        traintuples.append(traintuple)
+
+    spec = factory.create_aggregate_algo()
+    aggregate_algo = session.add_aggregate_algo(spec)
+
+    spec = factory.create_aggregatetuple(
+        algo=aggregate_algo,
+        objective=objective,
+        worker=session.node_id,
+        in_models_keys=traintuples,
```
jmorel: The arg is actually named `traintuples` in the factory.

tests/test_execution.py
```diff
@@ -245,3 +245,142 @@ def test_composite_traintuples_execution(factory, session):
...
+    test_data_sample = sessions[0].add_data_sample(spec)
+    # register objective on first node
+    spec = factory.create_objective(
+        dataset=datasets[0],
+        data_samples=[test_data_sample],
+    )
+    objective = sessions[0].add_objective(spec)
+
+    # register algos on first node
+    spec = factory.create_algo()
```
jmorel: Shouldn't we use create_composite_algo instead?

samlesu:
> LGTM! This will have to be properly tested once merged into `feature/generic-composite though`.

Yes! :)

PR #20, opened 2019-11-18
List and get compute plan + fix skaffold values

jmorel:
Companion PR to https://github.com/SubstraFoundation/substra-backend/pull/34 and https://github.com/SubstraFoundation/substra/pull/24

This PR adds tests for get/add/list compute plans and fixes existing tests (broken because of a skaffold conf update)


substratest/client.py
```diff
@@ -85,9 +85,34 @@ def add_testtuple(self, spec):
...
+        # The API outputs of the create and get/list methods differ at the
+        # moment. This harmonizes the returns so that they can be parsed.
+        # See:
+        # https://github.com/SubstraFoundation/substra-chaincode/issues/36
+        res['traintuples'] = res['traintupleKeys']
+        res['testtuples'] = res['testtupleKeys']
+        del res['traintupleKeys']
+        del res['testtupleKeys']
+        res['algoKey'] = spec.algo_key
+        res['objectiveKey'] = spec.objective_key
```
samlesu: I would prefer to create an `assets.ComputePlanCreated` structure instead. 

This will be also consistent with how we handle DataSample (different responses when doing read or write operations).

substratest/client.py
```diff
@@ -85,9 +85,34 @@ def add_testtuple(self, spec):
...
+        return [assets.ComputePlan.load(x) for x in res]
+
+    def get_compute_plan(self, *args, **kwargs):
+        res = self._client.get_compute_plan(*args, **kwargs)
+        compute_plan = assets.ComputePlan.load(res)
+        # the testtuples key of the compute_plan may be null instead
+        # of [] but should really be []
+        # See:
+        # https://github.com/SubstraFoundation/substra-chaincode/issues/36
+        compute_plan.testtuples = compute_plan.testtuples or []
```
samlesu: This should be done inside the `assets.py` module. We could maybe overwrite the `__init__` method or using `default_factory=list`?

tests/test_network.py
```diff
@@ -92,6 +92,28 @@ def test_list_nodes(network, session):
...
+    dataset = session.add_dataset(spec)
+
+    spec = factory.create_data_sample(test_only=False, datasets=[dataset])
+    data_sample = session.add_data_sample(spec)
+
+    spec = factory.create_compute_plan(algo=algo, objective=objective)
+    spec.add_traintuple(dataset=dataset,
+                        data_samples=[data_sample])
+
+    compute_plan = session.add_compute_plan(spec)
```
samlesu: I think we should always wait for a compute plan / traintuple / testtuple to be executed before ending the test.

tests/test_network.py
```diff
@@ -92,6 +92,28 @@ def test_list_nodes(network, session):
     assert set(network_node_ids).issubset(set(node_ids))
 
 
+def test_add_compute_plan(factory, session):
```
samlesu: Currently all tests executing stuff (and then "quite long" tests) are located inside the `test_execution*` files.

I think we should keep it that way to avoid having too many long tests. I think we should simply use and update an existing test. We could maybe rename this tests file to make things clearer or we could discuss about how organizing the tests.

substratest/client.py
jmorel: On it!

substratest/client.py
jmorel: Thanks for pointing out the `default_factory=list`, it gave me the occasion of looking into it. It doesn't solve the issue here because it only provides a default if no value is provided at all.

Here is an example:

```python
>>> @dataclass
... class C:
...     mylist: List[int] = field(default_factory=list)
...

>>> c = C()
>>> c.mylist == []
True
>>> c = C(mylist=None)
>>> c.mylist is None
True
```

jmorel: I found another solution btw, it's coming up soon :)

tests/test_network.py
jmorel: In this specific case, we're just testing that the create and get return the same values and that all sub-items were created. There are other tests that check that the execution went well, it's just not the purpose here.

tests/test_network.py
jmorel: (see discussion in the other comment)

jmorel:
@samlesu I've updated the tests with your suggestions, I'm waiting for the CI to finish but it should be ok.

jmorel:
I forgot there's no CI on this repository, @samlesu you can go ahead with the review.

PR #21, opened 2019-11-28
Generic compute plan

jmorel:
Add 2 tests for generic compute plans:
* make sure circular dependencies are handled correctly
* duplicate the `test_aggregate_composite_traintuples` in a compute plan

This required to update `factory.py` quite a bit.

### DIFF
--- a/substratest/assets.py
+++ b/substratest/assets.py
@@ -11,7 +11,17 @@ from . import errors
 FUTURE_TIMEOUT = 120  # seconds
 
 
-class Future:
+class BaseFuture(abc.ABC):
+    @abc.abstractmethod
+    def wait(self, timeout=FUTURE_TIMEOUT, raises=True):
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def get(self):
+        raise NotImplementedError
+
+
+class Future(BaseFuture):
     """Future asset."""
     # mapper from asset class name to client getter method
     _methods = {
@@ -50,43 +60,25 @@ class Future:
         return self._asset
 
 
-class ComputePlanFuture(Future):
-    _keys_properties = {
-        'ComputePlan': {
-            'traintuple_keys': 'traintuples',
-            'composite_traintuple_keys': 'composite_traintuples',
-            'aggregatetuple_keys': 'aggregatetuples',
-            'testtuple_keys': 'testtuples',
-        },
-        'ComputePlanCreated': {
-            'traintuple_keys': 'traintuple_keys',
-            'composite_traintuple_keys': 'composite_traintuple_keys',
-            'aggregatetuple_keys': 'aggregatetuple_keys',
-            'testtuple_keys': 'testtuple_keys',
-
-        },
-    }
-
+class ComputePlanFuture(BaseFuture):
     def __init__(self, asset, session):
         self._asset = asset
         self._getter = session.get_compute_plan
-        for k, v in enumerate(self._keys_properties[asset.__class__.name]):
-            setattr(self, f'_{k}', getattr(asset, v))
         self._get_traintuple = session.get_traintuple
         self._get_composite_traintuple = session.get_composite_traintuple
         self._get_aggregatetuple = session.get_aggregatetuple
         self._get_testtuple = session.get_testtuple
 
-    def wait(self, timeout=FUTURE_TIMEOUT, raises=True):
+    def wait(self, timeout=FUTURE_TIMEOUT):
         """wait until all tuples are completed (done or failed)."""
-        for key in self._traintuple_keys:
-            self._get_traintuple(key).future().wait(timeout, raises)
-        for key in self._composite_traintuple_keys:
-            self._get_composite_traintuple(key).future().wait(timeout, raises)
-        for key in self._aggregatetuple_keys:
-            self._get_aggregatetuple(key).future().wait(timeout, raises)
-        for key in self._testtuple_keys:
-            self._get_testtuple(key).future().wait(timeout, raises)
+        for key in self._asset.traintuple_keys:
+            self._get_traintuple(key).future().wait(timeout, raises=False)
+        for key in self._asset.composite_traintuple_keys:
+            self._get_composite_traintuple(key).future().wait(timeout, raises=False)
+        for key in self._asset.aggregatetuple_keys:
+            self._get_aggregatetuple(key).future().wait(timeout, raises=False)
+        for key in self._asset.testtuple_keys:
+            self._get_testtuple(key).future().wait(timeout, raises=False)
 
         return self.get()
 
@@ -95,6 +87,8 @@ class ComputePlanFuture(Future):
 
 
 class _FutureMixin(abc.ABC):
+    _future_cls = Future
+
     def attach(self, session):
         """Attach session to asset."""
         self._session = session
@@ -104,11 +98,11 @@ class _FutureMixin(abc.ABC):
         """Returns future from asset."""
         assert hasattr(self, 'status')
         assert hasattr(self, 'key')
+        return self._future_cls(self, self._session)
 
-        try:
-            return self.Meta.FutureCls(self, self._session)
-        except AttributeError:
-            return Future(self, self._session)
+
+class _ComputePlanFutureMixin(abc.ABC):
+    _future_cls = ComputePlanFuture
 
 
 def _convert(name):
@@ -376,26 +370,35 @@ class ComputePlanCreated(_Asset, _FutureMixin):
     aggregatetuple_keys: typing.List[str]
     testtuple_keys: typing.List[str]
 
-    class Meta:
-        FutureCls = ComputePlanFuture
-
 
 @dataclasses.dataclass
 class ComputePlan(_Asset):
     compute_plan_id: str
-    algo_key: str
     objective_key: str
     traintuples: typing.List[str]
     composite_traintuples: typing.List[str]
     aggregatetuples: typing.List[str]
     testtuples: typing.List[str]
 
-    class Meta:
-        FutureCls = ComputePlanFuture
-
     def __post_init__(self):
         if self.testtuples is None:
             self.testtuples = []
+
+    @property
+    def traintuple_keys(self):
+        return self.traintuples
+
+    @property
+    def composite_traintuple_keys(self):
+        return self.composite_traintuples
+
+    @property
+    def aggregatetuple_keys(self):
+        return self.aggregatetuples
+
+    @property
+    def testtuple_keys(self):
+        return self.testtuples
 
     def list_traintuples(self, session):
         return session.list_traintuples(filters=[f'traintuple:computePlanId:{self.compute_plan_id}'])
--- a/substratest/assets.py
+++ b/substratest/assets.py
@@ -61,29 +61,25 @@ class Future(BaseFuture):
 
 
 class ComputePlanFuture(BaseFuture):
-    def __init__(self, asset, session):
-        self._asset = asset
-        self._getter = session.get_compute_plan
-        self._get_traintuple = session.get_traintuple
-        self._get_composite_traintuple = session.get_composite_traintuple
-        self._get_aggregatetuple = session.get_aggregatetuple
-        self._get_testtuple = session.get_testtuple
+    def __init__(self, compute_plan, session):
+        self._compute_plan = compute_plan
+        self._session = session
 
     def wait(self, timeout=FUTURE_TIMEOUT):
         """wait until all tuples are completed (done or failed)."""
-        for key in self._asset.traintuple_keys:
-            self._get_traintuple(key).future().wait(timeout, raises=False)
-        for key in self._asset.composite_traintuple_keys:
-            self._get_composite_traintuple(key).future().wait(timeout, raises=False)
-        for key in self._asset.aggregatetuple_keys:
-            self._get_aggregatetuple(key).future().wait(timeout, raises=False)
-        for key in self._asset.testtuple_keys:
-            self._get_testtuple(key).future().wait(timeout, raises=False)
+        for key in self._compute_plan.traintuple_keys:
+            self._session.get_traintuple(key).future().wait(timeout, raises=False)
+        for key in self._compute_plan.composite_traintuple_keys:
+            self._session.get_composite_traintuple(key).future().wait(timeout, raises=False)
+        for key in self._compute_plan.aggregatetuple_keys:
+            self._session.get_aggregatetuple(key).future().wait(timeout, raises=False)
+        for key in self._compute_plan.testtuple_keys:
+            self._session.get_testtuple(key).future().wait(timeout, raises=False)
 
         return self.get()
 
     def get(self):
-        return self._getter(self._asset.compute_plan_id)
+        return self._session.get_compute_plan(self._compute_plan.compute_plan_id)
 
 
 class _BaseFutureMixin(abc.ABC):
--- a/substratest/assets.py
+++ b/substratest/assets.py
@@ -74,8 +74,8 @@ class ComputePlanFuture(BaseFuture):
         # testtuples do not have a rank attribute
         tuples += self._compute_plan.list_testtuple(self._session)
 
-        for tuple in tuples:
-            tuple.future().wait(timeout, raises=False)
+        for tuple_ in tuples:
+            tuple_.future().wait(timeout, raises=False)
 
         return self.get()
 
--- a/substratest/assets.py
+++ b/substratest/assets.py
@@ -50,6 +50,50 @@ class Future:
         return self._asset
 
 
+class ComputePlanFuture(Future):
+    _keys_properties = {
+        'ComputePlan': {
+            'traintuple_keys': 'traintuples',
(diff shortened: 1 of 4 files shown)


substratest/factory.py
```diff
@@ -142,6 +142,7 @@ class Permissions:
 
 
 DEFAULT_PERMISSIONS = Permissions(public=True, authorized_ids=[])
+DEFAULT_OUT_MODEL_PERMISSIONS = Permissions(public=False, authorized_ids=[])
```
inalgnu: shouldn't this have at least one authorized node? as is not public?

tests/test_execution_compute_plan.py
```diff
@@ -202,46 +208,168 @@ def test_compute_plan_single_session_failure(factory, session):
...
+        algo=algo,
+        data_samples=[data_sample]
+    )
+
+    traintuple_spec_1.in_models_ids.append(traintuple_spec_2.id)
+    traintuple_spec_2.in_models_ids.append(traintuple_spec_1.id)
+
+    # TODO make sur the creation is rejected
+    cp = session.add_compute_plan(cp_spec)
+    assert False
```
inalgnu: missing EOF

substratest/factory.py
jmorel: I believe passing an empty list will be interpreted as "owner only".

inalgnu: yes you are right, the backend will add the current node permission

substratest/assets.py
```diff
@@ -50,6 +50,50 @@ def get(self):
         return self._asset
 
 
+class ComputePlanFuture(Future):
```
samlesu: What do you think about creating an abstract base class `BaseFuture` with two abstract methods `get()` and `wait()`.

Here, it seems that `ComputePlanFuture` does not really need to inherit from `Future`.

substratest/assets.py
```diff
@@ -60,7 +104,11 @@ def future(self):
         """Returns future from asset."""
         assert hasattr(self, 'status')
         assert hasattr(self, 'key')
-        return Future(self, self._session)
+
+        try:
+            return self.Meta.FutureCls(self, self._session)
+        except AttributeError:
+            return Future(self, self._session)
```
samlesu: I think it will be simpler to create 2 classes: `_FutureMixin`, `_ComputePlanFutureMixin` instead of using meta attributes. For instance:
```python


class _FutureMixin(abc.ABC):
    _future_class = Future

    def attach(self, session):
        """Attach session to asset."""
        self._session = session
        return self

    def future(self):
        """Returns future from asset."""
        assert hasattr(self, 'status')
        assert hasattr(self, 'key')
        return self._future_class(self, self._session)

class _ComputePlanFutureMixin(_FutureMixin):
    _future_class = Future
```

substratest/assets.py
```diff
@@ -50,6 +50,50 @@ def get(self):
...
+        self._getter = session.get_compute_plan
+        for k, v in enumerate(self._keys_properties[asset.__class__.name]):
+            setattr(self, f'_{k}', getattr(asset, v))
+        self._get_traintuple = session.get_traintuple
+        self._get_composite_traintuple = session.get_composite_traintuple
+        self._get_aggregatetuple = session.get_aggregatetuple
+        self._get_testtuple = session.get_testtuple
+
+    def wait(self, timeout=FUTURE_TIMEOUT, raises=True):
+        """wait until all tuples are completed (done or failed)."""
```
samlesu: If `raises=True` it will wait until the first failure.

substratest/factory.py
```diff
@@ -276,26 +311,69 @@ def _get_keys(obj, field='key'):
...
-    algo_key: str
     objective_key: str
     traintuples: typing.List[ComputePlanTraintupleSpec]
+    composite_traintuples: typing.List[ComputePlanCompositeTraintupleSpec]
+    aggregatetuples: typing.List[ComputePlanAggregatetupleSpec]
     testtuples: typing.List[ComputePlanTesttupleSpec]
 
-    def add_traintuple(self, dataset, data_samples, traintuple_specs=None, tag=None):
-        traintuple_specs = traintuple_specs or []
+    def add_traintuple(self, algo, dataset, data_samples, in_models_tuples=None, tag=''):
```
samlesu: IMHO, `tuple_specs` describes better the input variable than `in_models_tuples` (even if it's not aligned with the substra library).

substratest/assets.py
```diff
@@ -50,6 +50,50 @@ def get(self):
...
+            'testtuple_keys': 'testtuples',
+        },
+        'ComputePlanCreated': {
+            'traintuple_keys': 'traintuple_keys',
+            'composite_traintuple_keys': 'composite_traintuple_keys',
+            'aggregatetuple_keys': 'aggregatetuple_keys',
+            'testtuple_keys': 'testtuple_keys',
+
+        },
+    }
```
samlesu: Even if it takes more line, I think setting these properties manually would by easier to read and to maintain. 

We could handle the different case in the constructor. For instance:
```python
self.traintuples_keys = asset.traintuples if hasattr(asset, 'traintuples') else asset.traintuple_keys 
```

One other solution would be to add the required properties to the `ComputePlan` class (`traintuple_keys`...).

What do you think?

substratest/assets.py
jmorel: Definitely!

substratest/assets.py
jmorel: Ok!

substratest/assets.py
jmorel: Good catch !

substratest/factory.py
jmorel: I understand what you mean but I find `tuple_specs` not clear enough (not clear that it is supposed to mach the in_models and not clear that we're talking about *ComputePlanSpecs here). 
What about `in_tuple_cp_specs`?

substratest/assets.py
jmorel: As discussed in the office, I went with the last solution.

substratest/assets.py
```diff
@@ -321,24 +363,57 @@ class Meta:
 
 
 @dataclasses.dataclass
-class ComputePlanCreated(_Asset):
+class ComputePlanCreated(_Asset, _FutureMixin):
```
samlesu:
```suggestion
class ComputePlanCreated(_Asset, _ComputePlanFutureMixin):
```

substratest/assets.py
```diff
@@ -321,24 +363,57 @@ class Meta:
...
     compute_plan_id: str
     traintuple_keys: typing.List[str]
+    composite_traintuple_keys: typing.List[str]
+    aggregatetuple_keys: typing.List[str]
     testtuple_keys: typing.List[str]
 
 
 @dataclasses.dataclass
 class ComputePlan(_Asset):
```
samlesu: missing `_ComputePlanFutureMixin` ?

substratest/factory.py
samlesu: I'm not really convinced with `in_tuple_cp_specs`: `cp` is not explicit and I find it too verbose. 

In your initial proposition, I think that `models` and `tuples` are redundant (as a model is the output of a tuple). What about `in_models` then? It would be consistent with the substra library even if I don't think it is the best name as it **is** a tuple.

tests/test_execution_compute_plan.py
```diff
@@ -22,33 +22,33 @@ def test_compute_plan(global_execution_env):
...
 
     # submit compute plan and wait for it to complete
-    cp = session_1.add_compute_plan(cp_spec)
+    cp = session_1.add_compute_plan(cp_spec).future().wait()
 
-    traintuples = [
-        session_1.get_traintuple(key).future().wait()
-        for key in cp.traintuple_keys
-    ]
+    traintuples = cp.list_traintuples(session_1)
```
samlesu: 👍

tests/test_execution_compute_plan.py
```diff
@@ -150,46 +143,170 @@ def test_compute_plan_single_session_failure(global_execution_env):
...
         data_samples=[data_sample_3],
-        traintuple_specs=[traintuple_spec_2]
+        in_models_tuples=[traintuple_spec_2]
     )
     cp_spec.add_testtuple(traintuple_spec_3)
 
     # Submit compute plan and wait for it to complete
-    cp = session.add_compute_plan(cp_spec)
+    cp_created = session.add_compute_plan(cp_spec)
+    cp = cp_created.future().wait()
```
samlesu: Could be done on a single line.
 We need also to add the option `raises=False` to ensure this function does not raise an exception as soon as the first tuple failed.

substratest/assets.py
samlesu: I haven't been very clear but the goal of my comment was to update the docstring and not to remove the wait option.

The goal of having a `raises` option is to ensure that by default the test will fail if a tuple execution failed. So I think we should keep it and pass it to the tuple `wait` methods.

tests/test_execution_compute_plan.py
```diff
@@ -150,46 +143,170 @@ def test_compute_plan_single_session_failure(global_execution_env):
...
-    compute_plan = session.get_compute_plan(cp.compute_plan_id)
-    assert cp.compute_plan_id == compute_plan.compute_plan_id
-    assert set(cp.traintuple_keys) == set(compute_plan.traintuples)
-    assert set(cp.testtuple_keys) == set(compute_plan.testtuples)
+    assert cp_created.compute_plan_id == cp.compute_plan_id
+    assert set(cp_created.traintuple_keys) == set(cp.traintuples)
+    assert set(cp_created.testtuple_keys) == set(cp.testtuples)
+
+
+def test_compute_plan_aggregate_composite_traintuples(factory, session_1, session_2):
```
samlesu: We could use the latest `execution_env` fixture.

tests/test_execution_compute_plan.py
```diff
@@ -150,46 +143,170 @@ def test_compute_plan_single_session_failure(global_execution_env):
...
+    traintuple_spec_2 = cp_spec.add_traintuple(
+        dataset=dataset,
+        algo=algo,
+        data_samples=[data_sample]
+    )
+
+    traintuple_spec_1.in_models_ids.append(traintuple_spec_2.id)
+    traintuple_spec_2.in_models_ids.append(traintuple_spec_1.id)
+
+    # TODO make sur the creation is rejected
```
samlesu: comment to be removed?
the `substra` library raising an `InvalidRequest` should be sufficient to ensure that the create has been rejected.

substratest/assets.py
jmorel: My reasoning was that we always want a compute plan to wait for all tuples to be over and not raise at the first one that fails. Otherwise we'll have potential pending/executing tuples at the end of the test that may prevent other tests from running successfully.

For example, let's say we have 1 traintuple and 10 composite_tuples, if the traintuple fails, the test would end immediately while there are still 10 ongoing composite tuples.

substratest/factory.py
jmorel: I agree it's not the best name, but at least it's consistent. This is something that ought to be changed throughout the platform. Let's go with in_models!

substratest/assets.py
jmorel: Yes 🤦‍♂

tests/test_execution_compute_plan.py
jmorel: I wanted to keep the cp as returned by the add_compute_plan function to make sure that the data returned by the add and get are the same ones (even if under different names). This is what the tests one lines 182-184 do.

substratest/assets.py
samlesu: That's indeed a very good reason!

tests/test_execution_compute_plan.py
samlesu: Yes, I haven't seen these lines. Sounds good to me.

tests/test_execution_compute_plan.py
jmorel: 👍

tests/test_execution_compute_plan.py
jmorel: Yes, it should have been removed when the pytest.raises was added.

tests/test_execution_compute_plan.py
```diff
@@ -150,46 +143,145 @@ def test_compute_plan_single_session_failure(global_execution_env):
...
+        data_samples=dataset.train_data_sample_keys
+    )
+
+    traintuple_spec_1.in_models_ids.append(traintuple_spec_2.id)
+    traintuple_spec_2.in_models_ids.append(traintuple_spec_1.id)
+
+    with pytest.raises(substra.exceptions.InvalidRequest) as e:
+        session.add_compute_plan(cp_spec)
+
+    assert 'circular' in str(e)
```
samlesu: Not sure this will work, cf pytest documentation for usage of `pytest.raises`:

```python
    with pytest.raises(RuntimeError) as excinfo:

        def f():
            f()

        f()
    assert "maximum recursion" in str(excinfo.value)
```

substratest/factory.py
```diff
@@ -276,26 +311,69 @@ def _get_keys(obj, field='key'):
...
+            algo_key=aggregate_algo.key,
+            worker=worker,
+            in_models_ids=[t.id for t in in_models],
+            tag=tag,
+        )
+        self.aggregatetuples.append(spec)
+        return spec
+
+    def add_composite_traintuple(self, composite_algo, dataset=None, data_samples=None,
+                                 in_head_model_tuple=None, in_trunk_model_tuple=None,
```
```suggestion
                                 in_head_model=None, in_trunk_model=None,
```

To be consistent in namings.

substratest/factory.py
```diff
@@ -276,26 +311,69 @@ def _get_keys(obj, field='key'):
...
+                                 out_trunk_model_permissions=None, tag=''):
```
```suggestion
                                 out_trunk_permissions=None, tag=''):
```

substratest/factory.py
jmorel: 👍

substratest/factory.py
jmorel: Actually in this case, the name used throughout the platform is out_trunk_model_permissions

samlesu: My bad, forget this comment.

substratest/assets.py
```diff
@@ -50,17 +60,60 @@ def get(self):
         return self._asset
 
 
-class _FutureMixin(abc.ABC):
+class ComputePlanFuture(BaseFuture):
+    def __init__(self, asset, session):
+        self._asset = asset
+        self._getter = session.get_compute_plan
```
samlesu: Just having a second thought about this:
- the `_asset` variable could be called directly `compute_plan`: it cannot be something else.
- the `_getter`, `_get_*` variables are not needed, we could directly store the input `session` and use it in the `wait()` and `get()` methods.

I think this code has been adapted from the already existing `Future` class and can be improved as it is dealing with only one type of asset.

tests/test_execution_compute_plan.py
AurelienGasser: Indeed it doesn't work ([anymore!](https://github.com/pytest-dev/pytest/issues/5579)). I used `str(e.value)` like in the example above and it works: https://github.com/SubstraFoundation/substra-tests/pull/21/commits/908a719e3137f83861997df5b5a9a4b89936b738

substratest/assets.py
```diff
@@ -50,17 +60,60 @@ def get(self):
...
+    def wait(self, timeout=FUTURE_TIMEOUT):
+        """wait until all tuples are completed (done or failed)."""
+        for key in self._asset.traintuple_keys:
+            self._get_traintuple(key).future().wait(timeout, raises=False)
+        for key in self._asset.composite_traintuple_keys:
+            self._get_composite_traintuple(key).future().wait(timeout, raises=False)
+        for key in self._asset.aggregatetuple_keys:
+            self._get_aggregatetuple(key).future().wait(timeout, raises=False)
+        for key in self._asset.testtuple_keys:
+            self._get_testtuple(key).future().wait(timeout, raises=False)
```
samlesu: The tuples should be ordered by execution order: it will avoid a timeout on the first one. To do that we could use the rank.

substratest/assets.py
jmorel: Done!

substratest/assets.py
```diff
@@ -67,14 +67,15 @@ def __init__(self, compute_plan, session):
...
-            self._session.get_testtuple(key).future().wait(timeout, raises=False)
+        tuples = (self._compute_plan.list_traintuple(self._session) +
+                  self._compute_plan.list_composite_traintuple(self._session) +
+                  self._compute_plan.list_aggregatetuple(self._session))
+        tuples = sorted(tuples, key=lambda t: t.rank)
+        # testtuples do not have a rank attribute
+        tuples += self._compute_plan.list_testtuple(self._session)
+
+        for tuple in tuples:
+            tuple.future().wait(timeout, raises=False)
```
samlesu: We should avoid using the reserved keyword tuple here. Could be replaced with `tuple_`?

jmorel: sure, my bad.

PR #22, opened 2019-12-02
Use data_network for execution tests

jmorel:
Since all execution tests rely on a very similar setup, we can use the existing (but dormant) `data_network` fixture so that base assets (datasets, data samples and objectives) are created only once.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -95,6 +95,10 @@ def data_network():
             spec = f.create_data_sample(datasets=[dataset], test_only=True)
             test_data_sample = sess.add_data_sample(spec)
 
+            # reload datasets (to ensure they are properly linked with the created data samples)
+            dataset = sess.get_dataset(dataset.key)
+            sess.state.datasets = [dataset]
+
             # create objective
             spec = f.create_objective(dataset=dataset, data_samples=[test_data_sample])
             sess.add_objective(spec)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -62,8 +62,8 @@ def network():
     return _get_network()
 
 
-@pytest.yield_fixture(scope='session')
-def data_network():
+@pytest.fixture(scope='session')
+def global_execution_env():
     """Network fixture with pre-existing assets in all nodes.
 
     The following asssets will be created for each node:
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -62,7 +62,7 @@ def network():
     return _get_network()
 
 
-@pytest.fixture(scope='session')
+@pytest.yield_fixture(scope='session')
 def data_network():
     """Network fixture with pre-existing assets in all nodes.
 
@@ -84,22 +84,22 @@ def data_network():
 
             # create dataset
             spec = f.create_dataset()
-            sess.add_dataset(spec)
+            dataset = sess.add_dataset(spec)
 
             # create train data samples
             for i in range(4):
-                spec = f.create_data_sample(test_only=False).fill_from_session(sess)
+                spec = f.create_data_sample(datasets=[dataset], test_only=False)
                 sess.add_data_sample(spec)
 
             # create test data sample
-            spec = f.create_data_sample(test_only=True).fill_from_session(sess)
-            sess.add_data_sample(spec)
+            spec = f.create_data_sample(datasets=[dataset], test_only=True)
+            test_data_sample = sess.add_data_sample(spec)
 
             # create objective
-            spec = f.create_objective().fill_from_session(sess)
+            spec = f.create_objective(dataset=dataset, data_samples=[test_data_sample])
             sess.add_objective(spec)
 
-        return f, n
+        yield f, n
 
 
 @pytest.fixture
--- a/tests/test_execution.py
+++ b/tests/test_execution.py
@@ -1,15 +1,14 @@
 import pytest
-import copy
 
 import substra
 
 import substratest as sbt
 
 
-def test_tuples_execution_on_same_node(data_network):
+def test_tuples_execution_on_same_node(global_execution_env):
     """Execution of a traintuple, a following testtuple and a following traintuple."""
-    factory, network = data_network
-    session = copy.deepcopy(network.sessions[0])
+    factory, network = global_execution_env
+    session = network.sessions[0].copy()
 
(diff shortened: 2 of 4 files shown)


tests/test_execution.py
```diff
@@ -5,22 +5,14 @@
 import substratest as sbt
 
 
-def test_tuples_execution_on_same_node(factory, session):
+def test_tuples_execution_on_same_node(data_network):
     """Execution of a traintuple, a following testtuple and a following traintuple."""
-    spec = factory.create_dataset()
-    dataset = session.add_dataset(spec)
+    factory, network = data_network
+    session = network.sessions[0]
```
samlesu: we should do a copy of the session here (to avoid updating the state of the data network session) ?

tests/conftest.py
```diff
@@ -62,7 +62,7 @@ def network():
     return _get_network()
 
 
-@pytest.fixture(scope='session')
+@pytest.yield_fixture(scope='session')
 def data_network():
```
samlesu: I think we can find a better name for this fixture as it is used now. Do you have any idea?

tests/test_execution.py
```diff
@@ -47,31 +39,23 @@ def test_tuples_execution_on_same_node(factory, session):
...
-    train_data_sample = session.add_data_sample(spec)
+    factory, network = data_network
+    session = network.sessions[0]
 
-    spec = factory.create_objective(
-        dataset=dataset,
-        data_samples=[test_data_sample],
-    )
-    objective = session.add_objective(spec)
+    # create test environment
```
samlesu:
```suggestion
    # get test environment
```

tests/test_execution.py
```diff
@@ -317,36 +272,20 @@ def test_aggregate_composite_traintuples(factory, session_1, session_2):
...
-        # register one data sample per dataset per round of aggregation
-        for _ in range(number_of_rounds):
-            spec = factory.create_data_sample(test_only=False, datasets=[dataset])
-            s.add_data_sample(spec)
     # reload datasets (to ensure they are properly linked with the created data samples)
+    datasets = sessions[0].state.datasets + sessions[1].state.datasets
     datasets = [
         sessions[i].get_dataset(d.key)
         for i, d in enumerate(list(datasets))
     ]
```
samlesu: This should be done in the `data_network` fixture.

tests/test_execution.py
```diff
@@ -5,22 +5,14 @@
...
-    train_data_sample = session.add_data_sample(spec)
-
-    spec = factory.create_objective(
-        dataset=dataset,
-        data_samples=[test_data_sample],
-    )
-    objective = session.add_objective(spec)
+    dataset = session.state.datasets[0]
+    objective = session.state.objectives[0]
+    train_data_samples = session.state.train_data_samples
```
samlesu: In all the tests we could remove this line and use directly `dataset.train_data_sample_keys`. Here we assume that all the train data samples in the session state belong to the same dataset.

tests/test_execution.py
jmorel: Good point, it's less risky that way!

tests/conftest.py
jmorel: Something to reflect its usage only in test_execution* files and its content ? What about `multi_node_execution_env` ?

tests/test_execution.py
jmorel: 👍

tests/test_execution.py
jmorel: 👍

tests/test_execution.py
jmorel: Well, they do :) But you're right, it'd be much more explicit. We'll just need to make sure the datasets do reference all the traintuple keys, and this is exactly the point of another of your comment. I'll make the change.

tests/conftest.py
samlesu: Sounds like a good idea. As it is the only execution environment, I will call it directly `execution_env` (or `global_execution_env`). `multi_node` could be confusing as it is also used in tests using a single node.

tests/conftest.py
```diff
@@ -62,7 +62,7 @@ def network():
     return _get_network()
 
 
-@pytest.fixture(scope='session')
+@pytest.yield_fixture(scope='session')
```
samlesu: according to pytest documentation we should not use `yield_fixture` (will be deprecated later). `fixture` also support yield statement (https://docs.pytest.org/en/latest/yieldfixture.html).

tests/test_execution.py
```diff
@@ -1,26 +1,18 @@
...
 import substratest as sbt
 
 
-def test_tuples_execution_on_same_node(factory, session):
+def test_tuples_execution_on_same_node(data_network):
     """Execution of a traintuple, a following testtuple and a following traintuple."""
-    spec = factory.create_dataset()
-    dataset = session.add_dataset(spec)
+    factory, network = data_network
+    session = copy.deepcopy(network.sessions[0])
```
samlesu: What about adding a `copy()` method to the `Session` class?

tests/conftest.py
jmorel: `global_execution_env` it is!

tests/test_execution.py
jmorel: It'd make things a bit easier, why not!

tests/test_execution.py
```diff
@@ -1,15 +1,14 @@
...
 import substratest as sbt
 
 
-def test_tuples_execution_on_same_node(data_network):
+def test_tuples_execution_on_same_node(global_execution_env):
     """Execution of a traintuple, a following testtuple and a following traintuple."""
-    factory, network = data_network
-    session = copy.deepcopy(network.sessions[0])
+    factory, network = global_execution_env
+    session = network.sessions[0].copy()
```
samlesu: I haven't thought about this earlier, sorry. The copy could be done directly inside the fixture: it will prevent fixture user to modify its content and thus simplify the fixture usage.

jmorel: This is a session fixture that is only executed once. Returning a copy in the fixture means that all tests will share the same copy. If we want to isolate the sessions during the tests, we'll have to make a copy in each test.

samlesu: Good point!

PR #23, opened 2019-12-03
Session state update on get

jmorel:
Update the state of the session at each get_*. Only assets that were added during the session are updated.

### DIFF
--- a/substratest/client.py
+++ b/substratest/client.py
@@ -10,10 +10,10 @@ DATASET_DOWNLOAD_FILENAME = 'opener.py'
 
 
 class _State:
-    """Session state."""
+    """Session state.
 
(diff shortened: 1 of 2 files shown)


substratest/client.py
```diff
@@ -10,10 +10,10 @@
 
 
 class _State:
-    """Session state."""
+    """Session state.
 
-    # TODO handle state updates when getting asset for instance; that would be
-    #      particularly useful for dataset update when adding a traintuple for instance
+    Represents all the asset that's been added during the life of the session.
```
samlesu:
```suggestion
    Represents all the assets that have been added during the life of the session.
```

samlesu:
LGTM

PR #25, opened 2019-12-06
Fail fast: remove evaluation for train like tuples

jmorel:
Companion PRs: 
* [backend](https://github.com/SubstraFoundation/substra-backend/pull/63)
* [substra](https://github.com/SubstraFoundation/substra/pull/52)

### DIFF
--- a/substratest/assets.py
+++ b/substratest/assets.py
@@ -264,9 +264,21 @@ class Objective(_Asset):
 
 
 @dataclasses.dataclass(frozen=True)
-class TupleDataset(_DataclassLoader):
+class TesttupleDataset(_DataclassLoader):
     key: str
     perf: float
+    keys: typing.List[str]
+    worker: str
+
+    class Meta:
+        mapper = {
+            'openerHash': 'key',
+        }
+
+
+@dataclasses.dataclass(frozen=True)
+class TraintupleDataset(_DataclassLoader):
+    key: str
     keys: typing.List[str]
     worker: str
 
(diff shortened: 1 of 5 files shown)


jmorel:
@samlesu I updated the new testtuple filter fixing the bug you fixed elsewhere in #31, could you re-review the PR?

samlesu:
> @samlesu I updated the new testtuple filter fixing the bug you fixed elsewhere in #31, could you re-review the PR?

Thanks, I updated the PR fixing the listing issue. I think it's good to keep related changes in the same PRs.

substratest/assets.py
```diff
@@ -264,7 +264,7 @@ class Objective(_Asset):
 
 
 @dataclasses.dataclass(frozen=True)
-class TupleDataset(_DataclassLoader):
+class TesttupleDataset(_DataclassLoader):
```
samlesu: minor and not 100% sure it would improve readability: but a TesttupleDataset is a TraintupleDataset with an additional perf field.

So we could have:
```python
class TesttupleDataset(TraintupleDataset):
    perf: float
```
```

samlesu: And we could keep the `TupleDataset` that is common for train like tuples and a dedicated `TesttupleDataset` inheriting from the `TupleDataset`.

jmorel: Since inheriting assets wasn't done anywhere I went with the duplication. I'll make the change.

samlesu: It's already done for Algo for instance.

samlesu: One other quick solution would be to make the perf field optional by adding a `= None` suffix.

jmorel: To recap our discussion in the office: the names `(Train/Test)tupleDataset` are ok because TraintupleDataset in used for Traintuples and CompositeTraintuples only.

Using Inheritance failed because of the __annotiations__ that are not enriched with the parent's class __annotations__.

In the end we went with the existing solution and will refactor this later on.

PR #26, opened 2019-12-06
Cancel compute plan

jmorel:
Companion PR:
- [substra](https://github.com/SubstraFoundation/substra/pull/56)
- [backend](https://github.com/SubstraFoundation/substra-backend/pull/65)
- [chaincode](https://github.com/SubstraFoundation/substra-chaincode/pull/53)
- [frontend](https://github.com/SubstraFoundation/substra-frontend/pull/33)

### DIFF
--- a/substratest/assets.py
+++ b/substratest/assets.py
@@ -21,12 +21,13 @@ class BaseFuture(abc.ABC):
         raise NotImplementedError
 
 
-class TupleStatus:
+class Status:
     doing = 'doing'
     done = 'done'
     failed = 'failed'
     todo = 'todo'
     waiting = 'waiting'
+    canceled = 'canceled'
 
 
 class Future(BaseFuture):
@@ -51,15 +52,19 @@ class Future(BaseFuture):
         """Wait until completed (done or failed)."""
         tstart = time.time()
         key = self._asset.key
-        while self._asset.status not in [TupleStatus.done, TupleStatus.failed]:
+        while self._asset.status not in [Status.done, Status.failed, Status.canceled]:
             if time.time() - tstart > timeout:
                 raise errors.FutureTimeoutError(f'Future timeout on {self._asset}')
 
             time.sleep(3)
             self._asset = self._getter(key)
 
-        if raises and self._asset.status == TupleStatus.failed:
+        if raises and self._asset.status == Status.failed:
             raise errors.FutureFailureError(f'Future execution failed on {self._asset}')
+
+        if raises and self._asset.status == Status.canceled:
+            raise errors.FutureFailureError(f'Future execution canceled on {self._asset}')
+
         return self.get()
 
     def get(self):
@@ -397,6 +402,7 @@ class Testtuple(_Asset, _FutureMixin):
 @dataclasses.dataclass
 class ComputePlan(_Asset, _ComputePlanFutureMixin):
     compute_plan_id: str
+    status: str
     traintuple_keys: typing.List[str]
     composite_traintuple_keys: typing.List[str]
     aggregatetuple_keys: typing.List[str]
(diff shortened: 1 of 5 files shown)


tests/test_execution_compute_plan.py
```diff
@@ -292,3 +292,34 @@ def test_compute_plan_circular_dependency_failure(global_execution_env):
...
+            dataset=dataset,
+            data_samples=[data_sample_key],
+            in_models=[previous_traintuple] if previous_traintuple else None
+        )
+
+    cp = session.add_compute_plan()
+    session.get_traintuple(cp.traintuple_keys[0]).future().wait()
+
+    cp = session.cancel_compute_plan(cp.compute_plan_id)
+    assert cp.status == 'cancelled'
```
inalgnu: we chose to use "canceled" 🇬🇧version, this will fail forever

tests/test_execution_compute_plan.py
```diff
@@ -292,3 +292,34 @@ def test_compute_plan_circular_dependency_failure(global_execution_env):
...
+    previous_traintuple = None
+    for data_sample_key in dataset.train_data_sample_keys:
+        previous_traintuple = cp_spec.add_traintuple(
+            algo=algo,
+            dataset=dataset,
+            data_samples=[data_sample_key],
+            in_models=[previous_traintuple] if previous_traintuple else None
+        )
+
+    cp = session.add_compute_plan()
```
maeldebon: The `add_compute_plan` method requires a "spec" argument. Maybe we could use `cp_spec` as it is done in the previous test?

jmorel: Done, thanks!

tests/test_execution_compute_plan.py
jmorel: Updated, thanks! I used @maeldebon's new status variables for that.

substratest/assets.py
```diff
@@ -385,6 +386,7 @@ class Meta:
 class ComputePlan(_Asset, _ComputePlanFutureMixin):
     compute_plan_id: str
     objective_key: str
+    status: str
```
samlesu: We could create a ComputePlan class, it would be more explicit and will be simpler to maintain.

If the status are exactly the same, we could simply do:

```python
class ComputePlanStatus(TupleStatus):
    pass
```

tests/test_execution_compute_plan.py
```diff
@@ -294,3 +294,35 @@ def test_compute_plan_circular_dependency_failure(global_execution_env):
...
+            data_samples=[data_sample_key],
+            in_models=[previous_traintuple] if previous_traintuple else None
+        )
+
+    cp = session.add_compute_plan(cp_spec)
+    first_traintuple = [t for t in cp.attach(session).list_traintuple() if t.rank == 0][0]
+    first_traintuple.future().wait()
+    cp = session.cancel_compute_plan(cp.compute_plan_id)
+
+    assert cp.status == assets.TupleStatus.canceled
```
samlesu: minor: as said in a previous comment, we should not compare a compute plan status with a tuple status. 

I see 2 options:
- have a `ComputePlanStatus` class
- rename `TupleStatus` to a more generic name: `FutureStatus` ?

jmorel: I like having a single class with a generic name but FutureStatus doesn't seem right. I really didn't see any issue with using TupleStatus though, especially since the compute plan status is inferred from its tuples' statuses. I'll create a ComputePlanStatus so that we can move forward.

samlesu: I agree, `FutureStatus` does not seem right. We could directly use `Status`, we don't have other statuses.

jmorel: 👍

PR #27, opened 2019-12-06
Add test for failed task retry

AurelienGasser:
Companion PR: https://github.com/SubstraFoundation/substra-backend/pull/66


AurelienGasser:
Sounds good!

The main reason I wrote this test is that it make it very convenient for testing (that the retry works). I wasn't sure myself if it was worth adding to the tests repo. With the new information from @inel (we're going to use a celery option instead of a homemade retry mechanism) it makes even less sense to add this test. Thank you!

Closing

Issue #29, opened 2019-12-09
Substratest: avoid using string to define tuple status


samlesu:
It is better to use global variables or custom type for this status. It improves code readability, code maintainability and is less error prone.

maeldebon:
Fixed in #30

PR #30, opened 2019-12-09
Avoid using strings to define tuple status

maeldebon:
As specified in the related issue, it is better to use global variables or custom type for this status. It improves code readability, code maintainability and is less error prone.

🔗 Related issue: #29

### DIFF
--- a/substratest/assets.py
+++ b/substratest/assets.py
@@ -21,12 +21,12 @@ class BaseFuture(abc.ABC):
         raise NotImplementedError
 
 
-class TupleStatus(enum.Enum):
-    doing = enum.auto()
-    done = enum.auto()
-    failed = enum.auto()
-    todo = enum.auto()
-    waiting = enum.auto()
+class TupleStatus:
+    doing = 'doing'
+    done = 'done'
+    failed = 'failed'
+    todo = 'todo'
+    waiting = 'waiting'
 
 
 class Future(BaseFuture):
@@ -51,14 +51,14 @@ class Future(BaseFuture):
         """Wait until completed (done or failed)."""
         tstart = time.time()
         key = self._asset.key
-        while self._asset.status not in [TupleStatus.done.name, TupleStatus.failed.name]:
+        while self._asset.status not in [TupleStatus.done, TupleStatus.failed]:
             if time.time() - tstart > timeout:
                 raise errors.FutureTimeoutError(f'Future timeout on {self._asset}')
 
             time.sleep(3)
             self._asset = self._getter(key)
 
-        if raises and self._asset.status == TupleStatus.failed.name:
+        if raises and self._asset.status == TupleStatus.failed:
             raise errors.FutureFailureError(f'Future execution failed on {self._asset}')
         return self.get()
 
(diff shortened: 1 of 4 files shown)


tests/test_execution.py
```diff
@@ -4,6 +4,9 @@
 
 import substratest as sbt
 
+status_done = 'done'
+status_failed = 'failed'
```
samlesu: What about creating an TupleStatus type in the `substratest.assets` module?
It could be an enum.

substratest/assets.py
```diff
@@ -1,5 +1,5 @@
 import abc
-import enum
+from enum import Enum, auto
```
samlesu: Why changing the import? I think using `enum.auto()` is more explicit than using `auto()` directly.

substratest/assets.py
```diff
@@ -21,6 +21,11 @@ def get(self):
         raise NotImplementedError
 
 
+class TupleStatus(Enum):
+    done = auto()
+    failed = auto()
```
samlesu: That's good! We could defined all the statuses here: 'todo', 'doing', 'waiting' and 'failed' (list to validate).

tests/test_execution.py
```diff
@@ -4,6 +4,9 @@
 
 import substratest as sbt
 
+status_done = sbt.client.assets.TupleStatus.done.name
+status_failed = sbt.client.assets.TupleStatus.failed.name
```
samlesu: Why creating global variables here? We could directly import the desired module. 

To avoid having a too long path we could import the assets module:
`from substratest import assets`

substratest/assets.py
maeldebon: Yes, you're right, I thought that it was cleaner this way

substratest/assets.py
maeldebon: Absolutely, I'll do that!

substratest/assets.py
```diff
@@ -21,9 +21,12 @@ def get(self):
...
 
-class TupleStatus(Enum):
-    done = auto()
-    failed = auto()
+class TupleStatus(enum.Enum):
+    doing = enum.auto()
+    done = enum.auto()
+    failed = enum.auto()
+    todo = enum.auto()
+    waiting = enum.auto()
```
jmorel: Very naive question here for @samlesu (since you suggested an enum) : what do we get out of an enum that we don't by using a basic class with string properties, something like:
```python
class TupleStatus:
   doing = 'doing'
```

In this specific case, using an enum creates a bit of overhead everywhere : we cant do `assert traintuple['status'] == TupleStatus.done`, we need to use `TupleStatus.done.name` which seems to defeat the purpose of using an enum in the first place.

samlesu: Good question !

I do agree that having to use the `name` property of an enum to check equality adds a layer of complexity. This could be improved by setting an enum in the `Asset` instance instead of a string. We should not forget that types defined in the dataclass are for "linting" purposes, there is no validation.

Right now, I don't see any advantage of using an enum rather than class properties. Till we don't have validation, using class properties seems like a good alternative solution.

PR #32, opened 2019-12-09
Asset ComputePlan: various improvements

samlesu:
- do not redefined `_FutureMixin.attach()`, already implemented in parent
  class
- `_FutureMixin`: raise a proper error if the asset has not been attached when using the session
- `ComputePlan.list_<tuple>`: use session set from the `_FutureMixin.attach` class method: 
- `ComputePlan.list_<tuple>`: add assert to ensure the list methods are working as expected
- test_compute_plan: small fix regarding typo in state attributes

### DIFF
--- a/substratest/assets.py
+++ b/substratest/assets.py
@@ -67,12 +67,13 @@ class ComputePlanFuture(BaseFuture):
 
     def wait(self, timeout=FUTURE_TIMEOUT):
         """wait until all tuples are completed (done or failed)."""
-        tuples = (self._compute_plan.list_traintuple(self._session) +
-                  self._compute_plan.list_composite_traintuple(self._session) +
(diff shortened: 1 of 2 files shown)


substratest/assets.py
```diff
@@ -88,21 +89,24 @@ class _BaseFutureMixin(abc.ABC):
...
         """Attach session to asset."""
-        self._session = session
+        self.__session = session
         return self
 
+    @property
+    def _session(self):
+        if not self.__session:
+            raise errors.TError(f'No session attached with {self}')
+        return self.__session
```
jmorel: This is very nice. I originally didn't want to directly use the attached session because it could not have been there, but this makes it very neat and explicit. 👍

samlesu:
Thanks!

PR #38, opened 2019-12-12
factory: add tests suite uuid to factory name

samlesu:
This will ensure all created assets in a test run will have a unique name. This could help when listing asset.


samlesu:
@jmorel could you please review this PR as you are using the asset generated names to do some checks on the CLI / frontend.

PR #40, opened 2019-12-13
Do not cancel  already done/failed/doing tuples

inalgnu:
related to https://github.com/SubstraFoundation/substra-chaincode/pull/62

### DIFF
--- a/tests/test_execution_compute_plan.py
+++ b/tests/test_execution_compute_plan.py
@@ -296,7 +296,7 @@ def test_compute_plan_circular_dependency_failure(global_execution_env):
     assert 'missing dependency among inModels IDs' in str(e.value)
 
 
-def test_execution_compute_plan_cancelled(global_execution_env):
+def test_execution_compute_plan_canceled(global_execution_env):
     factory, network = global_execution_env
     session = network.sessions[0].copy()
 
@@ -325,4 +325,7 @@ def test_execution_compute_plan_cancelled(global_execution_env):
 
     traintuples = cp.list_traintuple()
     for t in traintuples:
+        # first traintuple should be done, others traintuples should be canceled
+        if t.rank == 0:
+            continue
         assert t.status == assets.Status.canceled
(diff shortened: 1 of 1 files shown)


jmorel:
Instead of just making sure that assets are not in todo/waiting, what do you think of making the second traintuple last a few more seconds so that we are sure that the first tuple will end in status done, the second one is in status doing and all remaining ones are in status canceled?

tests/test_execution_compute_plan.py
```diff
@@ -325,7 +325,6 @@ def test_execution_compute_plan_canceled(global_execution_env):
 
     traintuples = cp.list_traintuple()
     for t in traintuples:
-        # first traintuple should be done, others traintuples should be canceled
-        if t.rank == 0:
-            continue
-        assert t.status == assets.Status.canceled
+        # first asset is done, second asset is doing, others should be canceled
+        # no traintuple should be todo or waiting.
+        assert t.status not in [assets.Status.todo, assets.Status.waiting]
```
samlesu: 💯

tests/test_execution_compute_plan.py
```diff
@@ -325,7 +325,6 @@ def test_execution_compute_plan_canceled(global_execution_env):
 
     traintuples = cp.list_traintuple()
```
samlesu: I think we should do a `cp.wait()` before the end of the test.

Unfortunately it is currently the responsibility of the test to return after all running tuples have been completed. This is required to ensure that a new test will start with the substra network ready to execute new tuples.

samlesu:
> Instead of just making sure that assets are not in todo/waiting, what do you think of making the second traintuple last a few more seconds so that we are sure that the first tuple will end in status done, the second one is in status doing and all remaining ones are in status canceled?

I don't think it is a good idea. In my opinion the tests should stay as simple as possible and as fast as possible. This is adding a bit of complexity and I don't think it is worth it.

tests/test_execution_compute_plan.py
thibaultrobert: Just to be sure you think it should be added  at the bottom at the test, after the assertions, right ?

samlesu: It could be added:
- before the `list_traintuples()` and then we should check the status are all in a final state canceled or done
- as the last line of the tests

jmorel:
Doing cp.wait() already checks that all tuples are done, so there's no need to add an explicit test that they are not in todo or doing.

PR #43, opened 2019-12-16
Fix compute plan test

### LINKED ISSUE #42: Failure on test tests/test_execution_compute_plan.py::test_compute_plan
Tests traceback:
```
        # check all traintuples are done and check they have been executed on the expected
        # node
        for t in traintuples:
            assert t.status == assets.Status.done

        traintuple_1, traintuple_2, traintuple_3 = traintuples

        assert len(traintuple_3.in_models) == 2

>       assert traintuple_1.dataset.worker == session_1.node_id
E       AssertionError: assert 'MyOrg2MSP' == 'MyOrg1MSP'
E         - MyOrg2MSP
E         ?      ^
E         + MyOrg1MSP
E         ?      ^
```

It could be the traintuples list not ordered as expected.

This is a random failure, I cannot reproduce it.

samlesu:
Fix #42 

Random failure due to random ordering of tuples with the same rank.

Thanks @jmorel for the investigation!

### DIFF
--- a/substratest/assets.py
+++ b/substratest/assets.py
@@ -424,6 +424,7 @@ class ComputePlan(_Asset, _ComputePlanFutureMixin):
         ]
         tuples = self._session.list_traintuple(filters=filters)
         assert len(tuples) == len(self.traintuple_keys)
+        assert set(self.traintuple_keys) == set([t.key for t in tuples])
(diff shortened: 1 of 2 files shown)


samlesu:
I always forgot about `{ }` syntax for initializing a set!
It could be confusing as this syntax is also commonly used to initialized a dict.

PR #44, opened 2019-12-16
add test for intermediate model removal

AurelienGasser:
Companion PR: https://github.com/SubstraFoundation/substra-backend/pull/92

### DIFF
--- a/tests/settings.py
+++ b/tests/settings.py
@@ -74,6 +74,7 @@ def _load_skaffold(path):
     return Settings(
         path=path,
         nodes=nodes,
+        # the "options" field is not supported when parsing skaffold file
     )
 
 
--- a/tests/settings.py
+++ b/tests/settings.py
@@ -25,8 +25,14 @@ class NodeCfg:
 
 
 @dataclasses.dataclass(frozen=True)
+class Options:
+    enable_intermediate_model_removal: bool
+
+
+@dataclasses.dataclass(frozen=True)
 class Settings:
     path: str
+    options: Options
     nodes: typing.List[NodeCfg] = dataclasses.field(default_factory=list)
 
 
@@ -42,6 +48,9 @@ def _load_yaml(path):
     return Settings(
         path=path,
         nodes=nodes,
+        options=Options(
+            enable_intermediate_model_removal=data['enable_intermediate_model_removal']
+        )
     )
 
 
@@ -97,7 +106,3 @@ load()
 
 
 MSP_IDS = [n.msp_id for n in _SETTINGS.nodes]
-
-# <Additional network features>
-ENABLE_INTERMEDIATE_MODEL_REMOVAL = os.getenv('ENABLE_INTERMEDIATE_MODEL_REMOVAL') == '1'
-# </End additional network features>
--- a/tests/settings.py
+++ b/tests/settings.py
@@ -26,6 +26,7 @@ class NodeCfg:
 
 @dataclasses.dataclass(frozen=True)
 class Options:
+    celery_task_max_retry: int
     enable_intermediate_model_removal: bool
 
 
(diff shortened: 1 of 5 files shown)


tests/settings.py
```diff
@@ -97,3 +97,7 @@ def load():
 
 
 MSP_IDS = [n.msp_id for n in _SETTINGS.nodes]
+
+# <Additional network features>
+ENABLE_INTERMEDIATE_MODEL_REMOVAL = os.getenv('ENABLE_INTERMEDIATE_MODEL_REMOVAL') == '1'
+# </End additional network features>
```
AurelienGasser: @samlesu What do you think if this pattern? I went for the simplest solution here. I'm open to your suggestions/guidance :)

values.yaml
```diff
@@ -1,3 +1,4 @@
+enable_intermediate_model_removal: False
```
jmorel: You may also need to add this to local-values.yaml.

tests/settings.py
```diff
@@ -42,6 +48,9 @@ def _load_yaml(path):
     return Settings(
         path=path,
         nodes=nodes,
+        options=Options(
+            enable_intermediate_model_removal=data['enable_intermediate_model_removal']
```
samlesu: I think we could also set it by default to False for skaffold file (`line 75`) and maybe add a comment like "option blabla is not supported when parsing skaffold file".

Or maybe directly create a separate PR to get rid of the network configuration defined from skaffold file? This is not very useful and I don't think it's worth it to maintain.

values.yaml
AurelienGasser: Will do, thank you!

AurelienGasser: Fixed: https://github.com/SubstraFoundation/substra-tests/pull/44/commits/23c97c74d6ad2a365535a1b2f8619f8fa8fdca99

tests/settings.py
AurelienGasser: Does this work? https://github.com/SubstraFoundation/substra-tests/pull/44/commits/f82d199a04bc4c85728f2d8f3a37a99aa8788283

samlesu: Not sure as there is no default value.
I will create a separate PR to remove the skaffold mode, it will solve this.

values.yaml
samlesu: what about moving this field under an options field?
It would reflect the Settings dataclass and we could load it easily: `Options(**data['options'])`.

AurelienGasser: Awesome, great idea ! https://github.com/SubstraFoundation/substra-tests/pull/44/commits/274dde387dce5034b0990d20ce51340003a09df4

I also added the `celery_task_max_retry` option while I was there

AurelienGasser:
I marked this PR as "ready for review" even though the feature hasn't been implemented in the backend yet: since the feature is disabled by default merging this PR shouldn't break anything. Possible issues with the test itself could be addressed in a later PR. The main benefit of merging this PR now would be to make the "options system" available in master.

tests/test_execution.py
```diff
@@ -315,6 +320,22 @@ def test_aggregate_composite_traintuples(global_execution_env):
...
+    dataset = session.state.datasets[0]
+    algo = session.add_algo(spec)
+
+    spec = factory.create_traintuple(
+        algo=algo,
+        dataset=dataset,
+        data_samples=dataset.train_data_sample_keys,
+    )
+    traintuple = session.add_traintuple(spec).future().wait()
+    assert traintuple.status == assets.Status.failed
```
samlesu: Ideally it would be better to try to do a request "as a backend" to get the deleted model.
This will be closer to what we want to test and would also check that this request is correctly handled when the model has been deleted. Here, we cannot know for sure the failure reason.

Unfortunately this cannot be done now as the username/password are not available in the settings files.

AurelienGasser: Good point, I'll add a comment to reflect this!

PR #47, opened 2019-12-18
Remove parsing of skaffold file

samlesu:
This mode is unused and it doesn't make sense to maintain it.
It's gonna be harder to support it when the network configuration will be more complex.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,9 +20,9 @@ You will need to install it thanks to the `pip` binary.
 
 # Run the tests
 
-The network configuration must be defined in a yaml file.
+The network configuration is described in a yaml file.
 
-Two configurations are currently available:
+Two configuration files are currently available:
 - `values.yaml` (default): for networks started with Kubernetes
 - `local-values.yaml`: for networks started with docker-compose, or started manually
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -20,9 +20,9 @@ You will need to install it thanks to the `pip` binary.
 
 # Run the tests
 
-## Using a network configuration file
+The network configuration must be defined in a yaml file.
```
AurelienGasser: - What about "is" instead of "must be"?
- What about "described" instead of "defined"?

README.md
```diff
@@ -20,9 +20,9 @@ You will need to install it thanks to the `pip` binary.
...
 
-This is the mode used by default. Two files are currently available:
+Two configurations are currently available:
```
AurelienGasser: What about "configuration files" instead of "configurations"?

README.md
samlesu: Thanks.

README.md
samlesu: Thanks.

samlesu:
> I agree with @AurelienGasser's comments regarding the wording, but it's otherwise good for me!

👍 
Thanks and done.

samlesu:
Rebased after recent merges

PR #49, opened 2019-12-18
Update federated learning test

samlesu:
It will allow to have a test running a traintuple on each node of the network.

This is closer to a real use case example.

Also checks that we cannot add twice the same tuple (this check is done in the populate).

### DIFF
--- a/tests/test_execution.py
+++ b/tests/test_execution.py
@@ -83,7 +83,7 @@ def test_federated_learning_workflow(global_execution_env):
         assert traintuple.status == assets.Status.done
         assert traintuple.out_model is not None
         assert traintuple.tag == 'foo'
-        assert traintuple.compute_plan_id is not None
+        assert traintuple.compute_plan_id   # check it is not None or ''
 
         rank += 1
 
--- a/tests/test_execution.py
+++ b/tests/test_execution.py
@@ -28,6 +28,10 @@ def test_tuples_execution_on_same_node(global_execution_env):
     assert traintuple.status == assets.Status.done
     assert traintuple.out_model is not None
 
+    # check we cannot add twice the same traintuple
+    with pytest.raises(substra.exceptions.AlreadyExists):
+        session.add_traintuple(spec)
+
     # create testtuple
     # don't create it before to avoid MVCC errors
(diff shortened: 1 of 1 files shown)


tests/test_execution.py
```diff
@@ -47,51 +51,41 @@ def test_tuples_execution_on_same_node(global_execution_env):
...
+    datasets = [s.state.datasets[0] for s in network.sessions]
+    # check there is one dataset per node in the network
+    assert set([d.owner for d in datasets]) == set([s.node_id for s in network.sessions])
+
+    # create 1 traintuple per dataset and chain them
+    traintuple = None
+    rank = 0
+    for dataset in datasets:
+        traintuples = [traintuple] if traintuple else []
+        compute_plan_id = traintuple.compute_plan_id if traintuple else None
```
jmorel: I think this will leave the compute_plan_id to either `None` or `''`, I'm not sure we should include it in the test.

samlesu: Yes, it will leave the compute plan id to `None`.
 That's already the case: it's default value of the `compute_plan_id` input argument in the `factory.create_traintuple(` method.

tests/test_execution.py
```diff
@@ -47,51 +51,41 @@ def test_tuples_execution_on_same_node(global_execution_env):
...
+            traintuples=traintuples,
+            tag='foo',
+            rank=rank,
+            compute_plan_id=compute_plan_id,
+        )
+        traintuple = session.add_traintuple(spec).future().wait()
+        assert traintuple.status == assets.Status.done
+        assert traintuple.out_model is not None
+        assert traintuple.tag == 'foo'
+        assert traintuple.compute_plan_id is not None
```
samlesu: we should also check here that the `compute_plan_id` is not an empty string.

tests/test_execution.py
jmorel: After discussion I realized that the behavior of the backend is to auto-generate a compute plan id as soon as a traintuple of rank 0 is created. This explains why there is a compute_plan_id after the first tuple creation. So my point is moot.

tests/test_execution.py
jmorel: 👍

PR #50, opened 2019-12-19
Load testing

jmorel:
A few tests to try injecting larger and larger compute plans into the platform and making sure the platform can handle them.

Use `pytest tests/test_load.py -rs -v --durations=0` to run these tests only.


tests/test_load.py
```diff
@@ -0,0 +1,163 @@
...
+    """
+    Shape of the compute plan:
+
+    Node A: traintuple 0 -> traintuple -> ... -> traintuple
+    Node B:              -> traintuple -> ... -> traintuple
+    """
+
+    factory, network = global_execution_env
+    session_1 = network.sessions[0].copy()
+    session_2 = network.sessions[0].copy()
```
samlesu:
```suggestion
    session_2 = network.sessions[1].copy()
```

tests/test_load.py
```diff
@@ -0,0 +1,163 @@
+import pytest
+
+from substratest import assets
+
+# COMPUTE_PLAN_SIZES = [10, 100, 1000, 10000]
+COMPUTE_PLAN_SIZES = [100]
+
+
+@pytest.mark.parametrize('compute_plan_size', COMPUTE_PLAN_SIZES)
+def test_load_single_node_1_branch(compute_plan_size, global_execution_env):
```
samlesu: This is not really clear what's the goal of this test.

The goal of this test is to check that we can create a compute plan with a "lot of tuples".
That's the reason why it is cancelled shortly after creation.

Strictly speaking, I'm not sure it's a load test. This test should not take long to complete.

tests/test_load.py
```diff
@@ -0,0 +1,163 @@
...
+
+    cp = session.cancel_compute_plan(cp.compute_plan_id)
+    assert cp.status == assets.Status.canceled
+
+    cp = cp.future().wait()
+    assert cp.status == assets.Status.canceled
+
+    first_tuple = cp.list_traintuple()[0]
+    assert first_tuple.rank == 0
+    assert first_tuple.status == assets.Status.done
```
samlesu: I think the goal of this test is only to check that compute plan can be added and that the first tuple has been executed properly.

What about replacing the selected lines with:
```python 
    cp = session.cancel_compute_plan(cp.compute_plan_id).future().wait()
    assert cp.status == assets.Status.canceled
```

It will make the test simpler to read and it will be easier to understand what is tested.

Same for the 3 other tests.

tests/test_load.py
```diff
@@ -0,0 +1,163 @@
...
+    """
+    Shape of the compute plan:
+
+    Node A: traintuple 0 --> traintuple --> aggregate --> traintuple --> aggregate ...
+    Node B:              \-> traintuple /             \-> traintuple /
+    """
+
+    factory, network = global_execution_env
+    session_1 = network.sessions[0].copy()
+    session_2 = network.sessions[0].copy()
```
samlesu:
```suggestion
    session_2 = network.sessions[1].copy()
```

tests/test_load.py
jmorel: Thank you! That was a big typo. I fixed it everywhere.

tests/test_load.py
jmorel: Absolutely!

jmorel:
Since this is not necessary anymore and since we don't plan on running it regularly there's no point in keeping the PR open. I'm closing it.

PR #51, opened 2019-12-19
Subtratest: split factory to improve code modularity

samlesu:
Don't be afraid with the diff, it's mostly splitting the factory.py files and moving pieces around. This will greatly help to handle different asset implementations (to solve #6)

The factory files has been converted to a `factory` sub package with the following components:
- factory: the class to create spec objects
- specs: the assets metadata specification
- experiment: the assets data structure definition

I have one main uncertainty: 
- the name of the entity grouping a consistant set of assets implementations. The name of the module is currently `experiment`. Other ideas? Requirements are: single word (not compound) and should not be confusing.

To add a new set of assets implementations, a new python module must be added in `experiments`. It must defines the following global variables:
```python
data_sample_generator = None

OPENER_SCRIPT = None
METRICS_SCRIPT = None
ALGO_SCRIPT = None
COMPOSITE_ALGO_SCRIPT = None
AGGREGATE_ALGO_SCRIPT = None

METRICS_DOCKERFILE = None
ALGO_DOCKERFILE = None
```

This is not ideal but it's simple and doesn't change the current structure. We could see in a following PR how we could improve it, ideally python code won't be stored in docstrings.


experiments/dummy.py
```diff
@@ -0,0 +1,121 @@
...
+        with open(path) as f:
+            return json.load(f)
+    def _save_model(self, model, path):
+        with open(path, 'w') as f:
+            return json.dump(model, f)
+if __name__ == '__main__':
+    tools.algo.execute(TestCompositeAlgo())
+"""
+
+INVALID_ALGO_SCRIPT = ALGO_SCRIPT.replace('train', 'naitr')
```
samlesu: to be removed.

experiments/dummy.py
```diff
@@ -0,0 +1,121 @@
...
+FROM substrafoundation/substra-tools:{DEFAULT_SUBSTRATOOLS_VERSION}
+COPY metrics.py .
+ENTRYPOINT ["python3", "metrics.py"]
+"""
+
+ALGO_DOCKERFILE = f"""
+FROM substrafoundation/substra-tools:{DEFAULT_SUBSTRATOOLS_VERSION}
+COPY algo.py .
+ENTRYPOINT ["python3", "algo.py"]
+"""
```
samlesu: Not a good idea define experiment as a module: it will be very hard to change one of the item for a single test.

samlesu:
Closed, won't be done now (not a priority).

PR #57, opened 2020-01-21
Fix tests that were broken with non local envs

jmorel:
In this PR:
- fix `test_merge_permissions_denied_process ` that is executed only on envs with at least 3 nodes and that hadn't been updated to reflect API changes
- update `celery_task_max_retry` option in `values.yaml`: rename it as `celery_task_max_retries` and make its value the same as the `CELERY_TASK_MAX_RETRIES` option in the backend
- disable shared path test on remote env with no shared path
- make `test_execution_retry_on_fail` work no matter what the `celery_task_max_retries` value is.

### DIFF
--- a/tests/test_permissions.py
+++ b/tests/test_permissions.py
@@ -4,7 +4,6 @@ import pytest
 
 from substratest.factory import Permissions
 from substratest import assets
-
 from . import settings
 
 MSP_IDS = settings.MSP_IDS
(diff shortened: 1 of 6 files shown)


tests/test_permissions.py
```diff
@@ -173,15 +179,16 @@ def test_merge_permissions_denied_process(factory, network):
...
 
         # failed to add testtuple from node 3
         spec = factory.create_testtuple(
             objective=objective_1,
-            traintuple_spec=traintuple_2,
+            traintuple=traintuple_2,
         )
 
-        with pytest.raises(substra.exceptions.AuthorizationError):
+        with pytest.raises(substra.exceptions.InvalidRequest) as e:
```
samlesu: I'm not sure this is a good change.

The expected exception should be `AuthorizationError` and not `InvalidRequest` here?

jmorel: You're right, and your fix in https://github.com/SubstraFoundation/substra-backend/pull/107 allowed me to revert my change. I also took the liberty to add a very basic permission test checking that you cannot create a tuple referencing assets you don't have the rights to.

jmorel: You're right, and your fix in SubstraFoundation/substra-backend#107 allowed me to revert my change. I also took the liberty to add a very basic permission test checking that you cannot create a tuple referencing assets you don't have the rights to.

jmorel:
This has been tested against https://github.com/SubstraFoundation/substra-backend/pull/107

Issue #37, opened 2019-12-10
Substratest: Asset objects: add validation


samlesu:
Currently asset objects are defined through dataclass instances with type hinting: **this is not checking that each field type is correct**.

We could implement a validation layer to ensure that the objects returned by the SDK are as expected. 

The first step is to do some search and propose a solution (look at mypy, pydantic, marshmallow, ...).

It should be easy to convert camelCase fields to snake_case fields with the chosen solution and to be able to rename some fields (as currently done with the Meta class for few assets).

samlesu:
Closed by #59

PR #59, opened 2020-01-21
Add data-type validation with Pydantic

maeldebon:
🔗 Related issue: #37 

This PR aims to add a **data-type validation** on substra-tests in order to be able to check the type of data returned by the [backend](http://github.com/SubstraFoundation/substra-backend) and to return an error is the type received is not the one expected.
We needed to find a library for this.

Some ideas of libraries have been explored:
• [Marshmallow](https://github.com/marshmallow-code/marshmallow)
• [Pydantic](https://github.com/samuelcolvin/pydantic) 
• [MYPY](https://github.com/python/mypy)

We tried to use Marshmallow first, but its wasn't compatible with how substra-tests is made. Indeed, the class instantiation wasn't what we expected and didn't work.
MYPY can be used, but not as a data-type validation tool as we wanted.

So we decided to use **Pydantic** instead of these two options.

### DIFF
--- a/substratest/assets.py
+++ b/substratest/assets.py
@@ -167,8 +167,8 @@ class _DataclassLoader(abc.ABC):
 
             if hasattr(attr_type, '__origin__') and attr_type.__origin__ is typing.Union \
                     and len(attr_type.__args__) == 2 and type(None) in attr_type.__args__:
-                attr_type = [arg for arg in attr_type.__args__ if arg
-                             is not type(None)][0]  # noqa: E721
+                attr_type = [arg for arg in attr_type.__args__
+                             if arg is not type(None)][0]  # noqa: E721
 
             # because typing.List doesn't work the same way as the other types, we have to check
             # if attr_type is a class before using issubclass()
--- a/substratest/assets.py
+++ b/substratest/assets.py
@@ -167,7 +167,8 @@ class _DataclassLoader(abc.ABC):
 
             if hasattr(attr_type, '__origin__') and attr_type.__origin__ is typing.Union \
                     and len(attr_type.__args__) == 2 and type(None) in attr_type.__args__:
-                attr_type = [arg for arg in attr_type.__args__ if arg is not type(None)][0]
+                attr_type = [arg for arg in attr_type.__args__ if arg
+                             is not type(None)][0]  # noqa: E721
 
             # because typing.List doesn't work the same way as the other types, we have to check
             # if attr_type is a class before using issubclass()
--- a/substratest/assets.py
+++ b/substratest/assets.py
@@ -6,6 +6,7 @@ import typing
 import pydantic
 
 from inspect import isclass
+
 from . import errors, cfg
 
 
@@ -123,8 +124,8 @@ class _DataclassLoader(abc.ABC):
             attr_name = mapper[k] if k in mapper else _convert(k)
             if attr_name not in cls.__annotations__:
                 continue
-            # handle nested dataclasses;
-            # FIXME does not work for list of nested dataclasses
+            # handle nested structures;
+            # FIXME does not work for list of nested structures
             attr_type = cls.__annotations__[attr_name]
             # because typing.List doesn't work the same way as the other types, we have to check
             # if attr_type is a class before using issubclass()
@@ -151,21 +152,21 @@ class _Asset(_InternalStruct, abc.ABC):
 
 
 class _BaseFutureAsset(_Asset):
-    __session: typing.Any = None
     _future_cls = None
 
     def attach(self, session):
         """Attach session to asset."""
+        # XXX because Pydantic doesn't support private fields, we have to use
+        # __getattribute__ and __setattr__ (https://github.com/samuelcolvin/pydantic/issues/655)
         object.__setattr__(self, '__session', session)
         return self
 
     @property
     def _session(self):
-        # because self.__session doesn't work properly with Pydantic, we have to use
-        # __getattribute__ and __setattr__ (https://docs.python.org/3/reference/datamodel.html)
-        if not object.__getattribute__(self, '__session'):
+        __session = object.__getattribute__(self, '__session')
+        if not __session:
             raise errors.TError(f'No session attached with {self}')
-        return object.__getattribute__(self, '__session')
+        return __session
 
     def future(self):
         """Returns future from asset."""
@@ -186,7 +187,26 @@ class _ComputePlanFutureAsset(_BaseFutureAsset):
     _future_cls = ComputePlanFuture
 
 
-class Permission(_InternalStruct):
+class _Frozen:
+    class Config:
+        allow_mutation = False
+
+
+class _FrozenAsset(_Asset, _Frozen):
+    """Add a Config class to an _Asset, which allows to set allow_mutation to False
+
+    Link to documentation: https://pydantic-docs.helpmanual.io/usage/models/#faux-immutability
+    """
+
+
+class _FrozenInternalStruct(_InternalStruct, _Frozen):
+    """Add a Config class to an _InternalStruct, which allows to set allow_mutation to False
+
+    Link to documentation: https://pydantic-docs.helpmanual.io/usage/models/#faux-immutability
+    """
+
+
+class Permission(_FrozenInternalStruct):
     public: bool
     authorized_ids: typing.List[str]
 
@@ -196,12 +216,12 @@ class Permission(_InternalStruct):
         }
 
 
-class Permissions(_InternalStruct):
+class Permissions(_FrozenInternalStruct):
     """Permissions structure stored in various asset types."""
     process: Permission
 
 
-class DataSampleCreated(_Asset):
+class DataSampleCreated(_FrozenAsset):
     key: str
     validated: bool
     path: str
@@ -212,15 +232,15 @@ class DataSampleCreated(_Asset):
         }
 
 
-class DataSample(_Asset):
+class DataSample(_FrozenAsset):
     key: str
     owner: str
     data_manager_keys: typing.List[str]
 
 
-class ObjectiveDataset(_InternalStruct):
-    dataset_key: str = None
-    data_sample_keys: typing.List[str] = None
+class ObjectiveDataset(_FrozenInternalStruct):
+    dataset_key: str
+    data_sample_keys: typing.List[str]
 
     class Meta:
         mapper = {
@@ -228,7 +248,7 @@ class ObjectiveDataset(_InternalStruct):
         }
 
 
-class Dataset(_Asset):
+class Dataset(_FrozenAsset):
     key: str
     name: str
     owner: str
@@ -238,7 +258,7 @@ class Dataset(_Asset):
     test_data_sample_keys: typing.List[str] = None
 
 
-class _Algo(_Asset):
+class _Algo(_FrozenAsset):
     key: str
     name: str
     owner: str
@@ -257,7 +277,7 @@ class CompositeAlgo(_Algo):
     pass
 
 
-class Objective(_Asset):
+class Objective(_FrozenAsset):
     key: str
     name: str
     owner: str
@@ -265,8 +285,8 @@ class Objective(_Asset):
     test_dataset: ObjectiveDataset
 
 
-class TesttupleDataset(_InternalStruct):
-    key: str = None
+class TesttupleDataset(_FrozenInternalStruct):
+    key: str
     perf: float
     keys: typing.List[str]
     worker: str
@@ -277,8 +297,8 @@ class TesttupleDataset(_InternalStruct):
         }
 
 
-class TraintupleDataset(_InternalStruct):
-    key: str = None
+class TraintupleDataset(_FrozenInternalStruct):
+    key: str
     keys: typing.List[str]
     worker: str
 
@@ -288,7 +308,7 @@ class TraintupleDataset(_InternalStruct):
         }
 
 
-class InModel(_InternalStruct):
+class InModel(_FrozenInternalStruct):
     key: str = None
     storage_address: str = None
 
@@ -298,9 +318,9 @@ class InModel(_InternalStruct):
         }
 
 
-class OutModel(_InternalStruct):
-    key: str = None
-    storage_address: str = None
+class OutModel(_FrozenInternalStruct):
+    key: str
+    storage_address: str
 
     class Meta:
         mapper = {
@@ -309,7 +329,7 @@ class OutModel(_InternalStruct):
 
 
 class Traintuple(_FutureAsset):
-    key: str = None
+    key: str
     creator: str
     status: str
     dataset: TraintupleDataset
@@ -328,7 +348,7 @@ class Traintuple(_FutureAsset):
 
 
 class Aggregatetuple(_FutureAsset):
-    key: str = None
+    key: str
     creator: str
     status: str
     worker: str
@@ -346,7 +366,7 @@ class Aggregatetuple(_FutureAsset):
         }
 
 
-class OutCompositeModel(_InternalStruct):
+class OutCompositeModel(_FrozenInternalStruct):
     permissions: Permissions
     out_model: OutModel = None
 
@@ -362,8 +382,8 @@ class CompositeTraintuple(_FutureAsset):
     log: str
     in_head_model: InModel = None
     in_trunk_model: InModel = None
-    out_head_model: OutCompositeModel = None
-    out_trunk_model: OutCompositeModel = None
+    out_head_model: OutCompositeModel
+    out_trunk_model: OutCompositeModel
 
     class Meta:
         mapper = {
@@ -372,7 +392,7 @@ class CompositeTraintuple(_FutureAsset):
 
 
 class Testtuple(_FutureAsset):
-    key: str = None
+    key: str
     creator: str
     status: str
     dataset: TesttupleDataset
@@ -388,7 +408,7 @@ class Testtuple(_FutureAsset):
 
 
 class ComputePlan(_ComputePlanFutureAsset):
-    compute_plan_id: str = None
+    compute_plan_id: str
     status: str
     traintuple_keys: typing.List[str] = None
     composite_traintuple_keys: typing.List[str] = None
@@ -396,21 +416,14 @@ class ComputePlan(_ComputePlanFutureAsset):
     testtuple_keys: typing.List[str] = None
     tag: str
 
-    def __post_init__(self):
-        if self.traintuple_keys is None:
-            self.traintuple_keys = []
-
-        if self.composite_traintuple_keys is None:
-            self.composite_traintuple_keys = []
-
-        if self.aggregatetuple_keys is None:
-            self.aggregatetuple_keys = []
-
-        if self.testtuple_keys is None:
-            self.testtuple_keys = []
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.traintuple_keys = self.traintuple_keys or []
+        self.composite_traintuple_keys = self.composite_traintuple_keys or []
+        self.aggregatetuple_keys = self.aggregatetuple_keys or []
+        self.testtuple_keys = self.testtuple_keys or []
 
     def list_traintuple(self):
-        self.__post_init__()
         filters = [
             f'traintuple:computePlanID:{self.compute_plan_id}',
         ]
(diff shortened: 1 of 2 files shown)


substratest/assets.py
```diff
@@ -1,10 +1,11 @@
 import abc
 import enum
-import dataclasses
 import re
 import time
 import typing
+import pydantic
 
+from inspect import isclass
```
samlesu: To follow PEP8, you must group the import into 3 groups (https://www.python.org/dev/peps/pep-0008/#imports):
- standard library
- 3rd party library
- internal library

substratest/assets.py
```diff
@@ -168,15 +139,54 @@ def load(cls, d):
...
+
+    def attach(self, session):
+        """Attach session to asset."""
+        object.__setattr__(self, '__session', session)
+        return self
+
+    @property
+    def _session(self):
+        # because self.__session doesn't work properly with Pydantic, we have to use
+        # __getattribute__ and __setattr__ (https://docs.python.org/3/reference/datamodel.html)
```
samlesu: maybe saying that "becausse Pydantic doesn't support private fields" is clearer than saying "self.__session doesn't work" ?

substratest/assets.py
```diff
@@ -168,15 +139,54 @@ def load(cls, d):
...
+        if not object.__getattribute__(self, '__session'):
+            raise errors.TError(f'No session attached with {self}')
+        return object.__getattribute__(self, '__session')
+
+    def future(self):
+        """Returns future from asset."""
+        return self._future_cls(self, object.__getattribute__(self, '__session'))
```
samlesu: we could use `self._session` here instead of `object.__getattribute__(self, '__session')`

substratest/assets.py
```diff
@@ -168,15 +139,54 @@ def load(cls, d):
...
+
+
+class _FutureAsset(_BaseFutureAsset):
+    """Represents a single task that is executed on the platform."""
+    _future_cls = Future
+
+    def future(self):
+        assert hasattr(self, 'status')
+        assert hasattr(self, 'key')
+        return super().future()
```
samlesu: not related to this PR directly but could be removed.

substratest/assets.py
```diff
@@ -204,25 +212,22 @@ class Meta:
...
     owner: str
     data_manager_keys: typing.List[str]
 
 
-@dataclasses.dataclass(frozen=True)
-class ObjectiveDataset(_DataclassLoader):
-    dataset_key: str
-    data_sample_keys: typing.List[str]
+class ObjectiveDataset(_InternalStruct):
+    dataset_key: str = None
```
samlesu: Are you sure None is required here? This comment applies to all the default None values. :)

substratest/assets.py
```diff
@@ -396,14 +387,13 @@ class Meta:
...
-    composite_traintuple_keys: typing.List[str]
-    aggregatetuple_keys: typing.List[str]
-    testtuple_keys: typing.List[str]
+    traintuple_keys: typing.List[str] = None
+    composite_traintuple_keys: typing.List[str] = None
+    aggregatetuple_keys: typing.List[str] = None
+    testtuple_keys: typing.List[str] = None
     tag: str
 
     def __post_init__(self):
```
samlesu: Could we get rid of the __post_init__ method?
It would be good to ensure the default value of the `<tuple>_keys` fields is an empty list and not a None value.

maeldebon:
> It looks like you forgot to include pydantic as a dependency. In this case, it should be added to the `install_requires` list in `setup.py`

You're absolutely right!
But this file only exists in the [substra](http://github.com/substraFoundation/substra) repository, I'm not sure about where I'm supposed to add it for this project

substratest/assets.py
maeldebon: Yes, I was actually checking the doc to be sure that I made it right, I'll change that

substratest/assets.py
maeldebon: Yes, I will change this as well, it isn't really clear indeed

substratest/assets.py
maeldebon: I just tried, it doesn't work :/

jmorel:
@maeldebon My bad, I got confused! For this project it should be added in `requirements.txt`!

maeldebon:
> @maeldebon My bad, I got confused! For this project it should be added in `requirements.txt`!

Thanks! I'll add it! 🙂

substratest/assets.py
maeldebon: I just checked, this isn't required for all the current `None` values, but for some of them it is. I'll remove the unneeded 🙂

substratest/assets.py
maeldebon: Should we do this in this PR instead of another one?

substratest/assets.py
maeldebon: I'll try to, but I'm not sure that it is possible

maeldebon: When I try to replace `None` by `[]`, I get this error:

```
E           pydantic.error_wrappers.ValidationError: 3 validation errors for ComputePlan
E           composite_traintuple_keys
E             none is not an allowed value (type=type_error.none.not_allowed)
E           aggregatetuple_keys
E             none is not an allowed value (type=type_error.none.not_allowed)
E           testtuple_keys
E             none is not an allowed value (type=type_error.none.not_allowed)
```

substratest/assets.py
```diff
@@ -168,15 +139,54 @@ def load(cls, d):
...
+
+
+class _Asset(_InternalStruct, abc.ABC):
     """Represents assets stored in the Substra framework.
 
     Convert a dict with camel case fields to a Dataclass.
     """
 
 
-@dataclasses.dataclass(frozen=True)
```
jmorel: I see that frozen has disappeared, did you try replacing it with pydantic faux immutability? https://pydantic-docs.helpmanual.io/usage/models/#faux-immutability

substratest/assets.py
jmorel: The issue is that the chaincode returns null instead of an empty list. I'm afraid we can't solve it here.
As for the default `[]` value, aren't we going to run into this classic issue?
```
In [1]: def foo(l=[]):
   ...:     l.append('foo')
   ...:     return l

In [2]: foo()
Out[2]: ['foo']

In [3]: foo()
Out[3]: ['foo', 'foo']
```

substratest/assets.py
```diff
@@ -420,6 +410,7 @@ def __post_init__(self):
             self.testtuple_keys = []
 
     def list_traintuple(self):
+        self.__post_init__()
```
jmorel: `__post_init__` is part of how dataclasses work. If we're not using dataclasses we shouldn't probably use this name and I really don't think we should call it manually, especially in such a call which isn't supposed to alter the object.

substratest/assets.py
samlesu: It could be done later in a separate PR. :)

substratest/assets.py
samlesu: Good point, thanks @jmorel. So I think it should be done manually in the constructor function (something similar to the dataclass `__post_init__` function).

substratest/assets.py
maeldebon: I just tried and it works, I'll add it! 🙂

substratest/assets.py
maeldebon: Alright!

substratest/assets.py
```diff
@@ -1,9 +1,11 @@
...
 import enum
-import dataclasses
 import re
 import time
 import typing
+import pydantic
+
+from inspect import isclass
 
 from . import errors, cfg
```
samlesu: The order is still incorrect:
- second group should include only pydantic
- third group should include only errors / cfg
- first group should include all the other libs (standard)

substratest/assets.py
```diff
@@ -168,15 +140,73 @@ def load(cls, d):
...
+    def attach(self, session):
+        """Attach session to asset."""
+        # XXX because Pydantic doesn't support private fields, we have to use
+        # __getattribute__ and __setattr__ (https://github.com/samuelcolvin/pydantic/issues/655)
+        object.__setattr__(self, '__session', session)
+        return self
+
+    @property
+    def _session(self):
+        __session = object.__getattribute__(self, '__session')
```
samlesu: Are you sure `__session = object.__getattribute__(self, '__session')` will return None / False if `attach` has not been called?

maeldebon: Hmm, it actually returns this:
`E       AttributeError: 'Traintuple' object has no attribute '__session'`

substratest/assets.py
maeldebon: Thanks! I just modified this 🙂

substratest/assets.py
```diff
@@ -186,7 +187,26 @@ class _ComputePlanFutureAsset(_BaseFutureAsset):
...
 
 
-class Permission(_InternalStruct):
+class _Frozen:
+    class Config:
+        allow_mutation = False
+
+
+class _FrozenAsset(_Asset, _Frozen):
+    """Add a Config class to an _Asset, which allows to set allow_mutation to False
```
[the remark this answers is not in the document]
maeldebon: Yes, it could indeed be better!

substratest/assets.py
samlesu: So it means that the condition inside the next if will always be False.

In this case you will have to use a try/catch syntax.

substratest/assets.py
```diff
@@ -152,31 +121,93 @@ def load(cls, d):
...
-            attr_name = mapper[k] if k in mapper else _convert(k)
-            if attr_name not in cls.__annotations__:
-                continue
-            # handle nested dataclasses;
-            # FIXME does not work for list of nested dataclasses
-            attr_type = cls.__annotations__[attr_name]
-            if dataclasses.is_dataclass(attr_type) and isinstance(v, dict):
-                v = attr_type.load(v)
-            kwargs[attr_name] = v
+            if v is not None:
```
samlesu: To minimize nesting, prefer the following syntax:
```python
if v is None:
    continue
```

Minimizing nesting helps code readability.
It also minimizes code changes when modifying the code and avoid having huge diff in git.

substratest/assets.py
```diff
@@ -152,31 +121,93 @@ def load(cls, d):
...
+                # handle nested structures;
+                attr_type = cls.__annotations__[attr_name]
+                # because typing.List doesn't work the same way as the other types, we have to check
+                # if attr_type is a class before using issubclass()
+                if isclass(attr_type) and issubclass(attr_type, _DataclassLoader) \
+                        and isinstance(v, dict):
+                    v = attr_type.load(v)
+
+                elif hasattr(attr_type, '__origin__') and attr_type.__origin__ is list:
+                    attr_args = attr_type.__args__
```
samlesu: could be named `list_value_types` (same for next variable). This name would help to understand what this variable reprensents.

substratest/assets.py
```diff
@@ -152,31 +121,93 @@ def load(cls, d):
...
+                if attr_name not in cls.__annotations__:
+                    continue
+                # handle nested structures;
+                attr_type = cls.__annotations__[attr_name]
+                # because typing.List doesn't work the same way as the other types, we have to check
+                # if attr_type is a class before using issubclass()
+                if isclass(attr_type) and issubclass(attr_type, _DataclassLoader) \
+                        and isinstance(v, dict):
+                    v = attr_type.load(v)
+
```
samlesu: a comment would help to explain what's the purpose of the following condition:
```python
# handle list of internal structures
```

substratest/assets.py
```diff
@@ -152,31 +121,93 @@ def load(cls, d):
...
+                elif hasattr(attr_type, '__origin__') and attr_type.__origin__ is list:
```
samlesu: we could add a check also on `v`: `and isinstance(v, list)`

substratest/assets.py
```diff
@@ -396,28 +408,21 @@ class Meta:
...
-            self.aggregatetuple_keys = []
-
-        if self.testtuple_keys is None:
-            self.testtuple_keys = []
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.traintuple_keys = self.traintuple_keys or []
+        self.composite_traintuple_keys = self.composite_traintuple_keys or []
+        self.aggregatetuple_keys = self.aggregatetuple_keys or []
+        self.testtuple_keys = self.testtuple_keys or []
```
jmorel: I think we could remove the `= None` by simply inversing the order of things in the `__init__`:

```python
def __init__(self, *args, **kwargs):
    kwargs['traintuple_keys'] = kwargs.get('traintuple_keys') or []
    kwargs['composite_traintuple_keys'] = kwargs.get('composite_traintuple_keys') or []
    kwargs['aggregatetuple_keys'] = kwargs.get('aggregatetuple_keys') or []
    kwargs['testtuple_keys'] = kwargs.get('testtuple_keys') or []
    super().__init__(*args, **kwargs)
```

samlesu: > I just discovered that pydantic allows to use a `@dataclass` decorator similar to the one from dataclasses: https://pydantic-docs.helpmanual.io/usage/dataclasses/ It looks like it'd be a bit simpler than these new classes inheriting from pydantic.basemodel. What was the benefit of using BaseModel over this decorator?

Using a dataclass decorator or inheriting from a  schema class are both good solutions to me. What would be interesting to know is, do the modifications of classes inheritance are required (removing Mixin classes and multiple inheritance)?

samlesu:
> I just discovered that pydantic allows to use a `@dataclass` decorator similar to the one from dataclasses: https://pydantic-docs.helpmanual.io/usage/dataclasses/ It looks like it'd be a bit simpler than these new classes inheriting from pydantic.basemodel. What was the benefit of using BaseModel over this decorator?

Using a dataclass decorator or inheriting from a  schema class are both good solutions to me.

What would be interesting to know is, do the modifications of classes inheritance are required (removing Mixin classes and multiple inheritance)?

substratest/assets.py
```diff
@@ -396,28 +404,21 @@ class Meta:
...
     compute_plan_id: str
     status: str
-    traintuple_keys: typing.List[str]
-    composite_traintuple_keys: typing.List[str]
-    aggregatetuple_keys: typing.List[str]
-    testtuple_keys: typing.List[str]
+    traintuple_keys: typing.List[str] = None
+    composite_traintuple_keys: typing.List[str] = None
+    aggregatetuple_keys: typing.List[str] = None
+    testtuple_keys: typing.List[str] = None
```
samlesu: I'm not 100% but @jmorel advised to remove the `= None` as well.

maeldebon: Indeed, I changed this :)

jmorel:
I spent a little time understanding how we're handling values that may be "null". 

Currently we're removing them entirely from the input data and then providing a default (equal to None) in our assets. It works well in our tests, but it also means that the following two JSONs would be equally valid : `{"foo": "foo", "bar": null}` and `{"foo": "foo"}`. 

This would fail to detect an API change where empty keys are not included anymore, which would probably break a lot of code.

I'd be in favor of using `typing.Optional` which is made specifically for this point : telling that the key must be there but that the value can be None.

I wasn't sure it would work so I made a test and with the following diff I was able to remove all None defaults and replace them with optional args where necessary.

```diff
diff --git a/substratest/assets.py b/substratest/assets.py
index c4032da..b54bee5 100644
--- a/substratest/assets.py
+++ b/substratest/assets.py
@@ -157,14 +157,17 @@ class _DataclassLoader(abc.ABC):
         mapper = cls.Meta.mapper
         kwargs = {}
         for k, v in d.items():
-            if v is None:
-                continue
-
             attr_name = mapper[k] if k in mapper else _convert(k)
             if attr_name not in cls.__annotations__:
                 continue
             # handle nested structures;
             attr_type = cls.__annotations__[attr_name]
+
+            # handle optional arguments
+            if hasattr(attr_type, '__origin__') and attr_type.__origin__ is typing.Union \
+                    and len(attr_type.__args__) == 2 and type(None) in attr_type.__args__:
+                attr_type = [arg for arg in attr_type.__args__ if arg is not type(None)][0]  # noqa: E721
+
             # because typing.List doesn't work the same way as the other types, we have to check
             # if attr_type is a class before using issubclass()
             if isclass(attr_type) and issubclass(attr_type, _DataclassLoader) \
@@ -251,6 +254,8 @@ class Dataset(_Asset, _Frozen):
     owner: str
     objective_key: str
     permissions: Permissions
+    # the JSON data returned by list_dataset doesn't include the following keys at all
+    # they are only included in the result of get_dataset
     train_data_sample_keys: typing.List[str] = None
     test_data_sample_keys: typing.List[str] = None

@@ -279,7 +284,7 @@ class Objective(_Asset, _Frozen):
     name: str
     owner: str
     permissions: Permissions
-    test_dataset: ObjectiveDataset = None
+    test_dataset: typing.Optional[ObjectiveDataset]


 class TesttupleDataset(_InternalStruct, _Frozen):
@@ -335,8 +340,8 @@ class Traintuple(_Asset, _FutureMixin):
     rank: int
     tag: str
     log: str
-    in_models: typing.List[InModel] = None
-    out_model: OutModel = None
+    in_models: typing.Optional[typing.List[InModel]]
+    out_model: typing.Optional[OutModel]

     class Meta:
         mapper = {
@@ -365,7 +370,7 @@ class Aggregatetuple(_Asset, _FutureMixin):

 class OutCompositeModel(_Asset, _Frozen):
     permissions: Permissions
-    out_model: OutModel = None
+    out_model: typing.Optional[OutModel]


 class CompositeTraintuple(_Asset, _FutureMixin):
@@ -377,8 +382,8 @@ class CompositeTraintuple(_Asset, _FutureMixin):
     rank: int
     tag: str
     log: str
-    in_head_model: InModel = None
-    in_trunk_model: InModel = None
+    in_head_model: typing.Optional[InModel]
+    in_trunk_model: typing.Optional[InModel]
     out_head_model: OutCompositeModel
     out_trunk_model: OutCompositeModel
```

samlesu:
@jmorel Great investigation. Makes sense to me and we should use the `typing.Optional` field. As you said it could help to detect API changes. 👍

substratest/assets.py
```diff
@@ -155,11 +160,32 @@ def load(cls, d):
...
-            # FIXME does not work for list of nested dataclasses
+            # handle nested structures;
             attr_type = cls.__annotations__[attr_name]
-            if dataclasses.is_dataclass(attr_type) and isinstance(v, dict):
+
+            # handle optional arguments
+
+            if hasattr(attr_type, '__origin__') and attr_type.__origin__ is typing.Union \
+                    and len(attr_type.__args__) == 2 and type(None) in attr_type.__args__:
+                # noqa: E721
```
jmorel: This is uneffective as it only prevents qa on this line.

substratest/assets.py
```diff
@@ -155,11 +160,32 @@ def load(cls, d):
...
+                attr_type = [arg for arg in attr_type.__args__ if arg is not type(None)][0]
+                # attr_type =
+                # [arg for arg in attr_type.__args__ if not isinstance(arg, type(None))][0]
```
jmorel: Extra comment that ought to be removed.

substratest/assets.py
```diff
@@ -168,15 +194,23 @@ def load(cls, d):
...
 
     Convert a dict with camel case fields to a Dataclass.
     """
 
 
-@dataclasses.dataclass(frozen=True)
-class Permission(_DataclassLoader):
+class _Frozen:
+    class Config:
+        allow_mutation = False
```
jmorel: It looks like this is not working as expected. I tried the following:

```python 
In [3]: n = Node(id='foo', is_current=False)

In [4]: n
Out[4]: Node(id='foo', is_current=False)

In [5]: n.id = 'bar'

In [6]: n.id
Out[6]: 'bar'
```

It should instead have raised `TypeError: "Node" is immutable and does not support item assignment`

maeldebon: Yes indeed. I'll look for a solution 👍

samlesu: I'm not sure this constraint is bringing value to the tests library.
Even if the goal of this PR is to keep the same perimeter, in my opinion, we can remove the immutable assets feature.

substratest/assets.py
```diff
@@ -167,7 +167,8 @@ def load(cls, d):
 
             if hasattr(attr_type, '__origin__') and attr_type.__origin__ is typing.Union \
                     and len(attr_type.__args__) == 2 and type(None) in attr_type.__args__:
-                attr_type = [arg for arg in attr_type.__args__ if arg is not type(None)][0]
+                attr_type = [arg for arg in attr_type.__args__ if arg
+                             is not type(None)][0]  # noqa: E721
```
samlesu: The line break seems unusual. A better line break to improve readability would be before the `if`.

maeldebon: Yes, you're right. I wasn't sure about this

samlesu:
\o/

PR #62, opened 2020-01-23
Add model permissions check

jmorel:
NB: the test doesn't pass at the moment, it's waiting for an upcoming fix.

### DIFF
--- a/tests/test_permissions.py
+++ b/tests/test_permissions.py
@@ -198,7 +198,7 @@ def test_permissions_denied_model_process(factory, network):
     traintuple_1 = session_1.add_traintuple(spec).future().wait()
 
     assert not traintuple_1.permissions.process.public
-    assert traintuple_1.permissions.process.authorized_ids == [MSP_IDS[0]]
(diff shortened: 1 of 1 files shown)


tests/test_permissions.py
```diff
@@ -162,6 +162,55 @@ def test_permissions_denied_process(factory, network):
...
+    spec = factory.create_traintuple(
+        algo=algo_1,
+        dataset=dataset_1,
+        data_samples=dataset_1.train_data_sample_keys,
+        tag='foo',
+    )
+    traintuple_1 = session_1.add_traintuple(spec).future().wait()
+
+    assert not traintuple_1.permissions.process.public
+    assert traintuple_1.permissions.process.authorized_ids == [MSP_IDS[0]]
```
samlesu:
```suggestion
    assert traintuple_1.permissions.process.authorized_ids == [session_1.node_id]
```

Avoid usage of MSP_ID global variable (when possible).

jmorel: Good point, thanks!

PR #63, opened 2020-01-27
Add a fast mode for tests

### LINKED ISSUE #41: Tests: add a fast mode
We could have 2 ways to run the tests:
- fast: for a quick development loop: runs only a few relevant tests (should run in less than X minutes)
- full: run all the tests

jmorel:
Fixes #41 

Creates a new "slow" marker for tests and adds a new `fast-test` target in Makefile that will only run "fast" tests.

With the current tests marked as slow, `make test` runs in 184s for 71 tests while `make fast-test` runs in 48s for 59 tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,5 +5,8 @@ pyclean:
 test: pyclean
 	pytest tests -rs -v --durations=0
 
+fast-test: pyclean
+	pytest tests -rs -v --durations=0 -m "not slow"
+
 install:
 	pip3 install -r requirements.txt
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@ pyclean:
 test: pyclean
 	pytest tests -rs -v --durations=0
 
-fast-test: pyclean
+test-minimal: pyclean
 	pytest tests -rs -v --durations=0 -m "not slow"
 
 install:
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ pip3 install -r requirements.txt
 The tests suite requires a Substra network up and running. The network can be started
 either with skaffold (Kubernetes), with docker-compose, or manually.
 
-The substra project is needed for running the tests.  
+The substra project is needed for running the tests.
 It can be found [here](https://github.com/SubstraFoundation/substra)
 
 You will need to install it thanks to the `pip` binary.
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -38,6 +38,18 @@ To run the tests using the provided `local-values.yaml` (or a custom config file
 SUBSTRA_TESTS_CONFIG_FILEPATH=local-values.yaml make test
 ```
 
+# Fast mode
+
+Since tests can take a long time to run, some of them as marked as slow. You can run the "fast" ones with:
```
inalgnu:
```suggestion
Since tests can take a long time to run, some of them are marked as slow. You can run the "fast" ones with:
```

samlesu:
> Great changes! 🐎
> 
> The only concern I would have is that the word "fast" could be misleading, because it doesn't convey the fact that a bunch of tests are essentially skipped. So it could be interpreted as a feature that runs the same tests, but faster (e.g. in parallel or something). For that reason, what would you think of changing the wording from "fast" to "minimal", e.g. `make fast-test` => `make test-minimal`?

Sounds like a good idea!

jmorel:
test-minimal is it then!

README.md
```diff
@@ -38,6 +38,18 @@ To run the tests using the provided `local-values.yaml` (or a custom config file
 SUBSTRA_TESTS_CONFIG_FILEPATH=local-values.yaml make test
 ```
 
+# Fast mode
```
AurelienGasser:
```suggestion
# Minimal mode
```

Makefile
```diff
@@ -5,5 +5,8 @@ pyclean:
 test: pyclean
 	pytest tests -rs -v --durations=0
 
+test-minimal: pyclean
+	pytest tests -rs -v --durations=0 -m "not slow"
```
AurelienGasser: What about a `MINIMAL=1` option instead of a separate rule? i.e.

```
make test
make test MINIMAL=1
```

(If you think it's a good idea, I can help with setting it up)

README.md
jmorel: Of course I missed something 🤦‍♂

Makefile
jmorel: Could you tell me what is the advantage of using `MINIMAL=1` ? I feel it would make the option less discoverable (since `make` won't list it) and a bit harder to type (no autocompletion).

AurelienGasser: Both your points make a lot of sense! I suggested using a parameter out of personal preference, but after your remarks I agree with you :)

Also, sorry for not writing "nit" in front of my comment!

PR #66, opened 2020-01-29
Fix: dataset can be linked with only one objective

samlesu:
Add more tests to cover link between dataset and objective

Companion PRs:
- [substra](https://github.com/SubstraFoundation/substra/pull/94)
- [backend](https://github.com/SubstraFoundation/substra-backend/pull/133)

### DIFF
--- a/substratest/client.py
+++ b/substratest/client.py
@@ -262,3 +262,8 @@ class Session:
         compute_plan = assets.ComputePlan.load(res).attach(self)
         self.state.update_compute_plan(compute_plan)
         return compute_plan
+
+    def link_dataset_with_objective(self, dataset, objective):
+        self._client.link_dataset_with_objective(dataset.key, objective.key)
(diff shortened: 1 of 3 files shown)


substratest/client.py
```diff
@@ -262,3 +262,8 @@ def cancel_compute_plan(self, *args, **kwargs):
         compute_plan = assets.ComputePlan.load(res).attach(self)
         self.state.update_compute_plan(compute_plan)
         return compute_plan
+
+    def link_dataset_with_objective(self, dataset, objective):
+        self._client.link_dataset_with_objective(dataset.key, objective.key)
+        # XXX do not return anything as currently the chaincode simply returns the
+        #     updated dataset key
```
AurelienGasser: Does "XXX" have a special meaning?

samlesu: I'm using XXX to flag code that requires attention and that does not need to be updated or fixed now.

Issue #14, opened 2019-11-06
Improve default algo / metrics / opener


samlesu:
The current implementation of the algo, the metrics and the opener are returning hard coded values that don't depend on the input values.

It doesn't allow to check that the output models / predictions and score are computed as expected.

It will be better if the implementation of these assets make sense from a Machine Learning point of view.

Tasks:
- [ ] Define structure to use (must be simple to implement and simple to use, i.e. expected results should be easy to compute on the side)
- [ ] Implement it
- [ ] Improve existing tests to ensure the task outputs are correct

jmorel:
I think that from the test framework, the only output we can see is the potential score. Which means in order to add checks on the execution, we'd have to add testtuples pretty much everywhere. This would make tests slower to run. Is this still a thing we want to do?

samlesu:
Yes, indeed currently the only output is the score.

In my point of view if the algo is validated in at least one test it would be enough.
If the model / data samples have not a valid format the train/test tuple will (should) not be able to run successfully.

Kelvin-M:
Is it still relevant @samlesu @jmorel  ?

samlesu:
In my opinion we can close it, this ticket is too old :)

PR #67, opened 2020-01-29
Improved asset code

jmorel:
Attempted fix of #14 

The idea behind the change is to make every sample and model an int:
- a data sample value is 2
- a dataset X value is the sum of the X value of each data samples (i.e. `2*len(folders)` by default)
- a dataset y value is the number of data samples
- an out model value is the sum of the value of all in models + X
- an aggregate value is the average of all in model values
- an out head model value is the sum of the value of all in head models + X
- an out trunk model value is the sum of the value of all in trunk models + y

A prediction is the value of the model (or head model + trunk model for composite algos)

The score is the difference between the prediction and actual y (ie. tuple model value - number of testtpules).

### DIFF
--- a/substratest/factory.py
+++ b/substratest/factory.py
@@ -18,18 +18,26 @@ DEFAULT_SUBSTRATOOLS_VERSION = '0.5.0'
 # TODO improve opener get_X/get_y methods
 # TODO improve metrics score method
 
-DEFAULT_OPENER_SCRIPT = """
+DEFAULT_OPENER_SCRIPT = f"""
+import csv
 import json
+import os
 import substratools as tools
 class TestOpener(tools.Opener):
     def get_X(self, folders):
-        return folders
+        res = 0
+        for folder in folders:
+            with open(os.path.join(folder, '{DEFAULT_DATA_SAMPLE_FILENAME}'), 'r') as f:
+                reader = csv.reader(f)
+                for row in reader:
+                    res += int(row[0])
+        return res
     def get_y(self, folders):
-        return folders
+        return len(folders)
     def fake_X(self):
-        return 'fakeX'
+        return 1
     def fake_y(self):
-        return 'fakey'
+        return 1
     def get_predictions(self, path):
         with open(path) as f:
             return json.load(f)
@@ -43,7 +51,7 @@ import json
 import substratools as tools
 class TestMetrics(tools.Metrics):
     def score(self, y_true, y_pred):
-        return 101
+        return y_pred -y_true
 if __name__ == '__main__':
     tools.metrics.execute(TestMetrics())
 """
@@ -53,9 +61,9 @@ import json
 import substratools as tools
 class TestAlgo(tools.Algo):
     def train(self, X, y, models, rank):
-        return [0, 1]
+        return sum(models) + X
     def predict(self, X, model):
-        return [0, 99]
+        return model
     def load_model(self, path):
         with open(path) as f:
             return json.load(f)
@@ -71,7 +79,7 @@ import json
 import substratools as tools
 class TestAggregateAlgo(tools.AggregateAlgo):
     def aggregate(self, models, rank):
-        return [0, 66]
+        return int(sum(models) / len(models))
     def load_model(self, path):
         with open(path) as f:
             return json.load(f)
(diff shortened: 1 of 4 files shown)


substratest/factory.py
```diff
@@ -18,18 +18,26 @@
 # TODO improve opener get_X/get_y methods
 # TODO improve metrics score method
```
samlesu: Could be removed now :)

substratest/factory.py
```diff
@@ -43,7 +51,7 @@ def save_predictions(self, y_pred, path):
 import substratools as tools
 class TestMetrics(tools.Metrics):
     def score(self, y_true, y_pred):
-        return 101
+        return y_pred -y_true
```
```suggestion
        return y_pred - y_true
```

Missing space

samlesu:
I'm wondering if we could implement in assets.py a function `compute_plan_get_score(compute_plan):` computing the expected score from the compute plan.

Could be funny to do in a separate PR ;)

jmorel:
Yes absolutely. I'll update the code.

jmorel:
@AurelienGasser @samlesu could you review this again? I updated the code with Aurélien's suggestion.

substratest/factory.py
```diff
@@ -15,21 +15,26 @@
...
 class TestOpener(tools.Opener):
     def get_X(self, folders):
-        return folders
+        res = []
+        for folder in folders:
+            with open(os.path.join(folder, '{DEFAULT_DATA_SAMPLE_FILENAME}'), 'r') as f:
+                reader = csv.reader(f)
+                for row in reader:
+                    res.append(int(row[0]))
+        return res
```
AurelienGasser: Just a note: `get_X` should return a matrix (not an array like I said in my previous comment!). But we're good here because the returned array can be considered as a matrix that only has 1 column.

substratest/factory.py
```diff
@@ -15,21 +15,26 @@
...
     def get_y(self, folders):
-        return folders
+        return len(folders)
     def fake_X(self):
-        return 'fakeX'
+        return [2]
     def fake_y(self):
-        return 'fakey'
+        return 1
```
AurelienGasser:
```suggestion
        return [1]
```

substratest/factory.py
```diff
@@ -53,9 +58,9 @@ def score(self, y_true, y_pred):
 import substratools as tools
 class TestAlgo(tools.Algo):
     def train(self, X, y, models, rank):
-        return [0, 1]
+        return sum(models) + sum(X)
```
AurelienGasser: It's not very clear from reading this line that we're returning a model. 

What about we return an actual dummy model here, like in the [substratools tests](https://github.com/SubstraFoundation/substra-tools/blob/master/tests/test_algo.py#L39)? i.e. `return {'value': 3}`

substratest/factory.py
```diff
@@ -89,9 +94,11 @@ def save_model(self, model, path):
 import substratools as tools
 class TestCompositeAlgo(tools.CompositeAlgo):
     def train(self, X, y, head_model, trunk_model, rank):
-        return [0, 1], [0, 2]
+        out_head_model = head_model + sum(X) if head_model else sum(X)
+        out_trunk_model = trunk_model + y if trunk_model else y
+        return out_head_model, out_trunk_model
```
AurelienGasser: Here we could return `out_head_model['value'] + out_trunk_model['value']`

substratest/factory.py
```diff
@@ -71,7 +76,7 @@ def save_model(self, model, path):
 import substratools as tools
 class TestAggregateAlgo(tools.AggregateAlgo):
     def aggregate(self, models, rank):
-        return [0, 66]
+        return int(sum(models) / len(models))
```
AurelienGasser: Here we could return the average `model['value']`

jmorel:
My issue here is that the more "real" the code, the less predictable the result. The overall goal was to validate that executions occurs and that data indeed flows between the different parts of the task, not to replicate a real-life use case. We have the examples for that and even so, real-life is so dependent on the use-case that anything we produce is always a bit off.

AurelienGasser:
Closed in favor of https://github.com/SubstraFoundation/substra-tests/pull/72

PR #69, opened 2020-02-03
Update cp

samlesu:
Companion PRs:
* https://github.com/SubstraFoundation/substra/pull/104
* https://github.com/SubstraFoundation/substra-backend/pull/137
* https://github.com/SubstraFoundation/substra-chaincode/pull/89

### DIFF
--- a/substratest/client.py
+++ b/substratest/client.py
@@ -150,6 +150,12 @@ class Session:
         self.state.compute_plans.append(compute_plan)
         return compute_plan
 
+    def update_compute_plan(self, spec, *args, **kwargs):
+        res = self._client.update_compute_plan(spec.to_dict(), *args, **kwargs)
+        compute_plan = assets.ComputePlan.load(res).attach(self)
+        self.state.compute_plans.append(compute_plan)
+        return compute_plan
+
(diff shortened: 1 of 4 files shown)


substratest/client.py
```diff
@@ -150,6 +150,12 @@ def add_compute_plan(self, spec, *args, **kwargs):
         self.state.compute_plans.append(compute_plan)
         return compute_plan
 
+    def update_compute_plan(self, spec, *args, **kwargs):
+        res = self._client.update_compute_plan(spec.compute_plan_id, spec.to_dict(), *args, **kwargs)
+        compute_plan = assets.ComputePlan.load(res).attach(self)
+        self.state.compute_plans.append(compute_plan)
```
jmorel: This shouldn't be an append but rather a replace I think.

samlesu: Good comment.
As discussed previously, the session state is not very usable right now and does not contain the last up-to-date assets.

As a short term fix we should remove the `append()`. We will see later how the state could be improved.

samlesu: Ups, to be consistent with existing code the best solution seems to call `update_compute_plan()`. The state will be reworked later.

PR #71, opened 2020-02-03
Update composite structure

Kelvin-M:
https://github.com/SubstraFoundation/substra-backend/pull/139
https://github.com/SubstraFoundation/substra-chaincode/pull/84

### DIFF
--- a/substratest/assets.py
+++ b/substratest/assets.py
@@ -363,7 +363,7 @@ class Aggregatetuple(_Asset, _FutureMixin):
 
 
 @dataclasses.dataclass(frozen=True)
-class OutCompositeModel(_DataclassLoader):
+class OutCompositeTrunkModel(_DataclassLoader):
     permissions: Permissions
     out_model: OutModel = None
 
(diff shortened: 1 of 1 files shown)


substratest/assets.py
```diff
@@ -370,7 +386,7 @@ class CompositeTraintuple(_Asset, _FutureMixin):
     log: str
     in_head_model: InModel = None
     in_trunk_model: InModel = None
-    out_head_model: OutCompositeModel = None
+    out_head_model: OutCompositeHeadModel = None
     out_trunk_model: OutCompositeModel = None
```
samlesu: The existing `OutCompositeModel` could be renamed `OutCompositeTrunkModel`? 

One other solution would have been to set the `OutModel.storage_address` field with a default `None` value (but the responses validation won't check if a "normal" OutModel has always a field `storage_address` set).

Kelvin-M: Yes for the renaming ! I'm waiting for other review on the companion PR to see what will be the final result.

As we want to consider Head Model as private (like data samples) having a storage address at `None` won't be the behavior that we want. We don't want to expose any storage address for this asset.

Kelvin-M: Done

PR #72, opened 2020-02-03
Improved asset code (with realistic interfaces)

AurelienGasser:
As discussed with @jmorel here is an alternative implementation for the opener/algo/metrics.

They are meant to 1. have an implementation that is very simple and 2. use interfaces that match the intended usage, in particular:

- `train` returns a model structure
- `predict` calculates `y` based on the input `model` and the input `X` 

Please let me know your thoughts. I'm particularly interested in feedback on the comments,

## Performance impact

I compared the time it takes to run `make test` between:

- `master`
- `improved_asset_code`
- `improved_asset_code_v2`

I ran `make test` 10x on each branch. Both `improved_asset_code` and `improved_asset_code_v2` take longer to run, but the increase is very small.

image

(those numbers are not worth much, it's just to get a general idea of the size of the increase)

In a future iteration, we could make changes so that the new algo/opener/metrics are only used for a few specific tests that ensure that data flows correctly between opener, algo and metrics. All the other tests could use a dummy algo that does nothing (like in `master`). This would shave a few seconds :man_shrugging:

### DIFF
--- a/substratest/factory.py
+++ b/substratest/factory.py
@@ -16,21 +16,32 @@ DEFAULT_DATA_SAMPLE_FILENAME = 'data.csv'
 
 DEFAULT_SUBSTRATOOLS_VERSION = '0.6.0'
 
-# TODO improve opener get_X/get_y methods
-# TODO improve metrics score method
-
-DEFAULT_OPENER_SCRIPT = """
+DEFAULT_OPENER_SCRIPT = f"""
+import csv
 import json
+import os
 import substratools as tools
 class TestOpener(tools.Opener):
     def get_X(self, folders):
-        return folders
+        res = []
+        for folder in folders:
+            with open(os.path.join(folder, '{DEFAULT_DATA_SAMPLE_FILENAME}'), 'r') as f:
+                reader = csv.reader(f)
+                for row in reader:
+                    res.append(int(row[0]))
+        return res  # returns a list of 1's
     def get_y(self, folders):
-        return folders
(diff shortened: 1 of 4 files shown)


samlesu:
@AurelienGasser could you review @jmorel's comment to move on this topic ?

AurelienGasser:
@jmorel I finally had a chance to look at this.

I made changes, what do you think?

substratest/factory.py
```diff
@@ -16,21 +16,32 @@
...
+        for folder in folders:
+            with open(os.path.join(folder, '{DEFAULT_DATA_SAMPLE_FILENAME}'), 'r') as f:
+                reader = csv.reader(f)
+                for row in reader:
+                    res.append(int(row[1]))
+        return res  # returns a list of 2's
+    def fake_X(self):
+        return [1]
+    def fake_y(self):
+        return [2]
```
jmorel, line 44: There is a new n_samples arg for the fake_X and fake_y method, but this means the code could become 
```python
    def fake_X(self, n_samples=1):
        return [1] * n_samples
    def fake_y(self, n_samples=1):
        return [2] * n_samples
```

jmorel:
I'm adding @Esadruhn as a reviewer because she also made an attempt at proper asset code.

substratest/factory.py
AurelienGasser: nice! done

samlesu:
💯

PR #73, opened 2020-02-04
Additional failure tests

jmorel:
We were only testing the failure behavior for traintuples. This adds tests for aggregatuples and composite traintuples as well.

### DIFF
--- a/tests/test_execution.py
+++ b/tests/test_execution.py
@@ -150,6 +150,64 @@ def test_traintuple_execution_failure(global_execution_env):
 
 
 @pytest.mark.slow
+def test_composite_traintuple_execution_failure(global_execution_env):
+    """Invalid composite algo script is causing traintuple failure."""
+    factory, network = global_execution_env
+    session = network.sessions[0].copy()
+
+    dataset = session.state.datasets[0]
+
+    spec = factory.create_composite_algo(py_script=sbt.factory.INVALID_COMPOSITE_ALGO_SCRIPT)
+    algo = session.add_composite_algo(spec)
+
+    spec = factory.create_composite_traintuple(
+        algo=algo,
+        dataset=dataset,
+        data_samples=dataset.train_data_sample_keys,
+    )
+    composite_traintuple = session.add_composite_traintuple(spec).future().wait(raises=False)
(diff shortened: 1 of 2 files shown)


tests/test_execution.py
```diff
@@ -149,6 +149,63 @@ def test_traintuple_execution_failure(global_execution_env):
...
+        composite_traintuples.append(session.add_composite_traintuple(spec))
+
+    spec = factory.create_aggregatetuple(
+        algo=aggregate_algo,
+        traintuples=composite_traintuples,
+        worker=session.node_id,
+    )
+    aggregatetuple = session.add_aggregatetuple(spec).future().wait(raises=False)
+    for composite_traintuple in composite_traintuples:
+        assert session.get_composite_traintuple(composite_traintuple.key).status == assets.Status.done
```
samlesu: Really minor: could be done in two lines.

jmorel: You mean no for loop and two assertions or just break the L204 into 2 lines?

samlesu: Sorry for the confusion,  just breaking the L204 into 2 line.
But that's so minor that I approved the PR, so up to you to do the change or not. :)

tests/test_execution.py
```diff
@@ -149,6 +149,64 @@ def test_traintuple_execution_failure(global_execution_env):
...
+
+    spec = factory.create_aggregatetuple(
+        algo=aggregate_algo,
+        traintuples=composite_traintuples,
+        worker=session.node_id,
+    )
+    aggregatetuple = session.add_aggregatetuple(spec).future().wait(raises=False)
+    for composite_traintuple in composite_traintuples:
+        composite_traintuple = session.get_composite_traintuple(composite_traintuple.key)
+        assert composite_traintuple.status == assets.Status.done
```
AurelienGasser: Minor: considering that the test's purpose is to ensure the aggtuple fails, do we need this block that ensures composite traintuples succeeded?

jmorel: If one of the composite traintuple failed, the aggregatuple would also automatically fail. This makes sure the reason the aggregatetuple failed is its invalid code and not an invalid dependency.

PR #75, opened 2020-02-05
Update state after get for all assets

jmorel:
The state wasn't updated after a get for algos (all 3 types) and objectives. This PR harmonizes the get behavior. 

It also updates the _update_assets method so that it either update the asset in the list if it already exists or inserts it if it doesn't.

### DIFF
--- a/substratest/client.py
+++ b/substratest/client.py
@@ -16,46 +16,66 @@ class _State:
     """
 
     def __init__(self):
-        self.datasets = []
-        self.test_data_samples = []
-        self.train_data_samples = []
-        self.objectives = []
-        self.algos = []
         self.aggregate_algos = []
-        self.composite_algos = []
-        self.traintuples = []
         self.aggregatetuples = []
+        self.algos = []
+        self.composite_algos = []
         self.composite_traintuples = []
-        self.testtuples = []
         self.compute_plans = []
+        self.datasets = []
+        self.objectives = []
+        self.test_data_samples = []
+        self.testtuples = []
+        self.train_data_samples = []
+        self.traintuples = []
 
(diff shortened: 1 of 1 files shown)


substratest/client.py
```diff
@@ -16,47 +16,67 @@ class _State:
...
+        found = False
+        for a in asset_list:
+            if asset.key == a.key:
+                found = True
+                new_asset_list.append(asset)
+            else:
+                new_asset_list.append(a)
+        if not found:
+            new_asset_list.append(asset)
+        setattr(self, asset_list_name, new_asset_list)
```
AurelienGasser: Is there a particular reason to change the implementation here?

AurelienGasser: Nevermind, I got it! 
The previous implementation didn't add missing elements.

samlesu: We could also use a dict to update the asset:

```python
existing_assets = {a.key: a for a in getattr(self, asset_list_name)}
existing_assets[asset.key] = asset
setattr(self, asset_list_name, list(existing_assets.values()))
```

Ideally, we should represent the assets using a dict (for performance reaon), but this could be done later.

jmorel:
So... 

I realized that this change introduces a potential risk: getting an asset owned by node_2 through session_1 will add the asset to session_1.state, thus breaking the underlying assumption that an asset in a session state is owned by the session.

@samlesu and I discussed it and the good solution here should probably be to have a global state shared by all sessions, with proxy methods allowing to access only the assets owned by each session. This way everybody could perform updates with no risk of contaminating each other with "falsy" information.

jmorel:
I'm closing this PR, implementing the fix proposed in #76 would be a much better solution.

PR #80, opened 2020-02-20
Fix asset loader (and restore py36 compatibility)

### LINKED ISSUE #78: Not compatible with python 3.6
It's working in python 3.7 but not in 3.6.

To reproduce:

Create a foo.yaml file with the following content: 
```yaml
algo:
  hash: a261b4558a46c0f8a5285b08316298609aaea29d0f30e1e2f51dfacaa27aa857
  name: cc6f5e17ce9542d3b41f02c9fbbefe38_global - Algo 0
  storageAddress: http://substra-backend.node-1.com/algo/a261b4558a46c0f8a5285b08316298609aaea29d0f30e1e2f51dfacaa27aa857/file/
computePlanID: ''
creator: MyOrg1MSP
dataset:
  keys:
  - 252a74f10f897b38a267bdc8a3d53d2a2738df689ecf244f046d046c43ba97f3
  - 4c6df43a64ced52e00b3fe702530ab1c2403d050dcc0939bf01d4149d15452dd
  - bf2329236d0e7156c464ed9bb3352af4dc5c0b929d9307fb4cb7771306140976
  - ec8c7f6e823cf46d49f066c7b74e30012b69c381213e0826935fa13d8d69bd3a
  openerHash: 109a87c1872558fcd3a01bc17e3b530c0de8ce592575fa06b5108b4162cee76b
  worker: MyOrg1MSP
inModels:
- hash: 72328140d2926398580faf7cf2322e851376d7f2ea5cc75a8e50c8e217695c0b
  storageAddress: http://substra-backend.node-1.com/model/72328140d2926398580faf7cf2322e851376d7f2ea5cc75a8e50c8e217695c0b/file/
  traintupleKey: 283e9ccd74bccb12413d50ad9035dcae7732e74601acd86da7dc0d23d3cbb7bf
key: 6ef5f35a529f6af84162163b90f7df4f1203ed23cc202a730e5140411d553249
log: ''
outModel:
  hash: 13b9e485072de1e36ff575661b52160231393ff23122172cd544fcac44ca5d31
  storageAddress: http://substra-backend.node-1.com/model/13b9e485072de1e36ff575661b52160231393ff23122172cd544fcac44ca5d31/file/
permissions:
  process:
    authorizedIDs: []
    public: true
rank: 0
status: done
tag: ''
```

Create a foo.py file with the following content:
```python
import yaml

import substratest as sbt

with open('foo.yaml') as f:
    d = yaml.load(f)

a = sbt.assets.Traintuple.load(d)
print(a)
```

Create the following Dockerfile:
```Dockerfile
FROM python:3.6

RUN pip install pyyaml substra

ADD requirements.txt .
RUN pip install -r requirements.txt


COPY substratest substratest/

ADD foo.yaml .
ADD foo.py .

ENV PYTHONPATH .

RUN python3 foo.py
```

Run the following command:
```
docker build -t substratest .
```

Once this is fixed, we will have to update the main README.

samlesu:
1. make asset loader py36 compatible:

The asset loader method was too complicated. Using private fields was not a good idea and should not be done.

The proposed solution is less flexible but easier to read, understand and to maintain.

Applying a set of fixed rules to rename the assets fields should not be issue. We expect the number of rules to decrease over time (and not the opposite!).

2. asset: remove openerHash to key mapper
3. asset: remove hash to key mapper

Close #78

### DIFF
--- a/substratest/assets.py
+++ b/substratest/assets.py
@@ -132,7 +132,6 @@ class _ComputePlanFutureMixin(_BaseFutureMixin):
 _MANUAL_CONVERTION_FIELDS = {
     'authorizedIDs': 'authorized_ids',
     'dataManagerKey': 'dataset_key',
-    'hash': 'key',  # TODO is it really needed?
     'pkhash': 'key',
 }
 
@@ -258,17 +257,17 @@ class TraintupleDataset(_InternalStruct):
 
 
 class InModel(_InternalStruct):
-    key: str
+    hash_: str = pydantic.Field(..., alias='hash')
     storage_address: str
 
 
 class OutModel(_InternalStruct):
-    key: str
+    hash_: str = pydantic.Field(..., alias='hash')
     storage_address: str
 
 
 class OutHeadModel(_InternalStruct):
-    key: str
+    hash_: str = pydantic.Field(..., alias='hash')
 
 
 class Traintuple(_Asset, _FutureMixin):
--- a/substratest/assets.py
+++ b/substratest/assets.py
@@ -133,7 +133,6 @@ _MANUAL_CONVERTION_FIELDS = {
     'authorizedIDs': 'authorized_ids',
     'dataManagerKey': 'dataset_key',
     'hash': 'key',  # TODO is it really needed?
-    'openerHash': 'key',  # TODO is it really needed?
     'pkhash': 'key',
 }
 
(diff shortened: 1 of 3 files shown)


substratest/assets.py
```diff
@@ -131,69 +129,26 @@ class _ComputePlanFutureMixin(_BaseFutureMixin):
...
+    """Converts asset camel case fields to snake case fields.
 
-    Provides a load method to create the object from a dictionary.
-    Converts automatically camel case fields to snake case fields and provides a mapper
-    to define custom fields mapping.
+    Special cases are handled through a global dict variable.
     """
+    # XXX using a mapper for converting specific is not very flexible as it will be
+    #     applied to all fields from all assets.
+    mapper = _MANUAL_CONVERTION_FIELDS
```
maeldebon: Because _MANUAL_CONVERTION_FIELDS is only used here, maybe we could move its content directly in `mapper`?

samlesu: It's a good practise to define **constant** variables as global variables.

samlesu:
Thanks for the review @maeldebon @AurelienGasser

substratest/assets.py
maeldebon: Okay 👍

PR #83, opened 2020-02-26
fix cp test

AurelienGasser:
This fixes the test `test_compute_plan` following [this chaincode PR](https://github.com/SubstraFoundation/substra-chaincode/pull/102).

`queryComputePlan` never returns any item in `IDToKey`, see [`getOutComputePlan`](https://github.com/SubstraFoundation/substra-chaincode/blob/34e90747562c9b42ebea22f08345a67cd5f2b2a5/chaincode/compute_plan.go#L280).

Alternatively, instead of this PR, we could change the chaincode behavior so that we return the entire set of id-to-key mappings. But then we'd risk hitting the gRPC size limit (which if I understood correctly was the whole point of the chaincode PR!)

### DIFF
--- a/tests/test_execution_compute_plan.py
+++ b/tests/test_execution_compute_plan.py
@@ -53,7 +53,10 @@ def test_compute_plan(global_execution_env):
     )
 
     # submit compute plan and wait for it to complete
-    cp = session_1.add_compute_plan(cp_spec).future().wait()
+    cp_added = session_1.add_compute_plan(cp_spec)
+    id_to_key = cp_added.id_to_key
+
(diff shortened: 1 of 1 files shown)


jmorel:
The code indeed fixes the test following the change in the chaincode but I'm not sure about the behavior. I created a small issue https://github.com/SubstraFoundation/substra-chaincode/issues/104 maybe we could discuss it there before merging this?

samlesu:
Hi @jmorel, I merged his PR to ensure the tests are not broken on the latest master branches.

Thanks for creating the issue, we should continue the discussion there.

Issue #76, opened 2020-02-12
Split session object into client and state


jmorel:
Currently each session has its own state which contains the assets created (and owned) by the session during the current test run. 

The objective of the session is to contain all assets necessary for the execution of a test. For example, the fixture `global_execution_env` returns sessions which contain all necessary assets to run `test_aggregatetuple`

In order to better reflect this, we'd like to separate the session object into 2 independent object:
* `Client` (replaces `Session`) is a wrapper around `substra.Client` to handle serialization
* `State` is an independent object with private properties (`state._dataset`) and public methods to access assets owned by a given node (`state.get_dataset(node_id)`)

Fixtures such as `global_execution_env` would then not return a factory and a network but instead a factory and a state containing all assets created by the fixture.

Individual tests such as `test_aggregatetuple` would therefore need to get as args the fixture `global_execution_env` and a `clients` fixture (or multiple individual fixtures `client_1`, `client_2` etc.)

This will lead to a pretty big renaming throughout the tests (no more sessions, only clients and states instead)

jmorel:
Thanks a lot @maeldebon for this very nice change! 

I see now a few follow up tasks:
* rename `Session` in `Client`
* remove `test_data_samples` and `test_data_samples` from `_Assets` since they are not used anywhere
* add a `filter_by(node_id)` method to the Assets object that accepts as argument a string (or even better, a string or a Client object) and returns a new Assets object which contains all datasets/objectives whose owner is the given node_id
* update all tests to use this new filter_by method instead of list comprehensions
* make _Assets a standard class (it does not need to depend on pydantic) and store actual assets in private properties (i.e. `_objectives`, `_datasets`) and have public getters (using `@property`) for them. This way the object will be effectively read-only.

maeldebon:
I'll do it @jmorel, thanks! 😄 
I was just checking for the following point:
> remove test_data_samples and test_data_samples from _Assets since they are not used anywhere

It looks like this is still used in some tests 🙁

jmorel:
Could you tell me where? I made a search for `test_data_samples` and `train_data_samples` but couldn't find any use of the variables in tests, only in `global_execution_env` where they are written into but never read from.

maeldebon:
> Could you tell me where? I made a search for `test_data_samples` and `train_data_samples` but couldn't find any use of the variables in tests, only in `global_execution_env` where they are written into but never read from.

[...]

PR #84, opened 2020-03-03
Split session object into client and state (2)

maeldebon:
🔗 **This Pull Request is related to the following issue: #76** 

As specified by @jmorel, there is still some work remaining on this point. He suggested the following changes:

- [x] Rename Session in Client
- [x] Remove `test_data_samples` and `train_data_samples` from Assets since they are not used anywhere
- [x] Add a `filter_by(node_id)` method to the Assets object that accepts as argument a string (or even better, a string or a Client object) and returns a new Assets object which contains all datasets/objectives whose owner is the given `node_id`
- [x] Update all tests to use this new `filter_by` method instead of list comprehensions
- [x] Make _Assets a standard class (it does not need to depend on Pydantic) and store actual assets in private properties (i.e. _objectives, _datasets) and have public getters (using @property) for them. This way the object will be effectively read-only.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -57,13 +57,13 @@ class _TestAssets:
 @dataclasses.dataclass
 class Network:
     options: settings.Options
-    sessions: typing.List[sbt.Client] = dataclasses.field(default_factory=list)
+    clients: typing.List[sbt.Client] = dataclasses.field(default_factory=list)
 
 
 def _get_network():
     """Create network instance from settings."""
     cfg = settings.load()
-    sessions = [sbt.Client(
+    clients = [sbt.Client(
         node_name=n.name,
         node_id=n.msp_id,
         address=n.address,
@@ -72,7 +72,7 @@ def _get_network():
     ) for n in cfg.nodes]
     return Network(
         options=cfg.options,
-        sessions=sessions,
+        clients=clients,
     )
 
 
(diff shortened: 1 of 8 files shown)


tests/conftest.py
```diff
@@ -117,22 +129,20 @@ def global_execution_env():
...
 
             # create test data sample
             spec = f.create_data_sample(datasets=[dataset], test_only=True)
             test_data_sample = sess.add_data_sample(spec)
-            assets.test_data_samples.append(test_data_sample)
 
             # reload datasets (to ensure they are properly linked with the created data samples)
             dataset = sess.get_dataset(dataset.key)
-            assets.datasets.append(dataset)
+            assets._datasets.append(dataset)
```
jmorel: I didn't think this all the way through it seems. I'm not very fond of access private properties directly. Maybe we could have an `add_dataset()` method? It's not the most elegant though. @samlesu any thought on the proper pattern to use here?

tests/conftest.py
```diff
@@ -117,22 +129,20 @@ def global_execution_env():
             # create train data samples
             for i in range(4):
                 spec = f.create_data_sample(datasets=[dataset], test_only=False)
-                data_sample = sess.add_data_sample(spec)
-                assets.train_data_samples.append(data_sample)
+                sess.add_data_sample(spec)
```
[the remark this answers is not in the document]
jmorel: This is true throughout the tests: `session` variable should be renamed in `client` I think.

tests/conftest.py
samlesu: IMO, it's ok to access private attributes of a class inside its own module. 
It would be indeed better to avoid accessing them outside of the class itself.

To do so, we could use the _TestsAssets constructor to set them inside the `global_execution_env` fixture. Something like:
```python
datasets = []
for ...
    dataset = ...
    datasets.append(dataset)
assets = _TestAssets(datasets=datasets)
```

As the goal is to ensure the tests cannot modify the `_TestAssets` instances, I think it would be a bad idea to implement setter methods.

jmorel: Didn't think of that, this is very clean indeed!

tests/conftest.py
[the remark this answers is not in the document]
maeldebon: Yes indeed, It'll be cleaner this way

PR #85, opened 2020-03-04
Pass factory, client & assets directly in arguments

maeldebon:
Currently, there are a few lines at the beginning of each test which allows to get the following variables:
• `factory` - *used for the `create_asset` methods*
• `initial_assets` - *initial assets used in the test*
• `network` - *used to get the client(s)*
They are set using this line:
```py
factory, initial_assets, network = global_execution_env
```  

The goal of this PR would be to pass directly these variable into the test, instead of getting them into. It avoids duplicated code and makes it easier to read.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -150,6 +150,67 @@ def global_execution_env():
 
 
 @pytest.fixture
+def initial_assets():
+    """Fixture with pre-existing assets in all nodes.
+
+    The following asssets will be created for each node:
+    - 4 train data samples
+    - 1 test data sample
+    - 1 dataset
+    - 1 objective
+
+    Network must started outside of the tests environment and the network is kept
+    alive while running all tests.
+
+    Returns the assets created.
+    """
+    network = _get_network()
+    factory_name = f"{TESTS_RUN_UUID}_global"
+
+    with sbt.AssetsFactory(name=factory_name) as f:
+        datasets = []
+        objectives = []
+        for client in network.clients:
+
+            # create dataset
+            spec = f.create_dataset()
+            dataset = client.add_dataset(spec)
+
+            # create train data samples
+            for i in range(4):
+                spec = f.create_data_sample(datasets=[dataset], test_only=False)
+                client.add_data_sample(spec)
+
+            # create test data sample
+            spec = f.create_data_sample(datasets=[dataset], test_only=True)
+            test_data_sample = client.add_data_sample(spec)
+
+            # reload datasets (to ensure they are properly linked with the created data samples)
+            dataset = client.get_dataset(dataset.key)
+            datasets.append(dataset)
+
+            # create objective
+            spec = f.create_objective(dataset=dataset, data_samples=[test_data_sample])
+            objective = client.add_objective(spec)
+            objectives.append(objective)
+
+        assets = _TestAssets(datasets=datasets, objectives=objectives)
+        return assets
+
+
+@pytest.fixture
+def initial_assets_1(initial_assets, client_1):
+    """Assets fixture (first node)."""
+    return initial_assets.filter_by(client_1.node_id)
+
+
+@pytest.fixture
+def initial_assets_2(initial_assets, client_2):
+    """Assets fixture (second node)."""
+    return initial_assets.filter_by(client_2.node_id)
+
+
+@pytest.fixture
 def client_1(network):
     """Client fixture (first node)."""
     return network.clients[0]
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -100,55 +100,6 @@ def network():
     return _get_network()
 
 
-# @pytest.fixture(scope='session')
-# def global_execution_env():
-#     """Network fixture with pre-existing assets in all nodes.
-#
-#     The following asssets will be created for each node:
-#     - 4 train data samples
-#     - 1 test data sample
-#     - 1 dataset
-#     - 1 objective
-#
-#     Network must started outside of the tests environment and the network is kept
-#     alive while running all tests.
-#
-#     Returns a tuple (factory, assets, Network).
-#     """
-#     n = _get_network()
-#     factory_name = f"{TESTS_RUN_UUID}_global"
-#
-#     with sbt.AssetsFactory(name=factory_name) as f:
-#         datasets = []
-#         objectives = []
-#         for client in n.clients:
-#
-#             # create dataset
-#             spec = f.create_dataset()
-#             dataset = client.add_dataset(spec)
-#
-#             # create train data samples
-#             for i in range(4):
-#                 spec = f.create_data_sample(datasets=[dataset], test_only=False)
-#                 client.add_data_sample(spec)
-#
-#             # create test data sample
-#             spec = f.create_data_sample(datasets=[dataset], test_only=True)
-#             test_data_sample = client.add_data_sample(spec)
-#
-#             # reload datasets (to ensure they are properly linked with the created data samples)
-#             dataset = client.get_dataset(dataset.key)
-#             datasets.append(dataset)
-#
-#             # create objective
-#             spec = f.create_objective(dataset=dataset, data_samples=[test_data_sample])
-#             objective = client.add_objective(spec)
-#             objectives.append(objective)
-#
-#         assets = _TestAssets(datasets=datasets, objectives=objectives)
-#         yield f, assets, n
-
-
 @pytest.fixture(scope="session")
 def data_envs():
     """Fixture with pre-existing assets in all nodes.
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -162,18 +162,6 @@ def data_env_2(default_data_env, client_2):
 
 
 @pytest.fixture
-def default_dataset(data_env_1):
-    """Fixture with pre-existing dataset in first node."""
-    return data_env_1.datasets[0]
-
-
-@pytest.fixture
-def default_objective(data_env_1):
-    """Fixture with pre-existing objective in first node."""
-    return data_env_1.objectives[0]
-
-
-@pytest.fixture
 def default_dataset_1(data_env_1):
     """Fixture with pre-existing dataset in first node."""
     return data_env_1.datasets[0]
@@ -195,6 +183,18 @@ def default_dataset_2(data_env_2):
 def default_objective_2(data_env_2):
     """Fixture with pre-existing objective in second node."""
     return data_env_2.objectives[0]
+
+
+@pytest.fixture
+def default_dataset(default_dataset_1):
+    """Fixture with pre-existing dataset in first node."""
+    return default_dataset_1
+
+
+@pytest.fixture
+def default_objective(default_objective_1):
+    """Fixture with pre-existing objective in first node."""
+    return default_objective_1
 
 
 @pytest.fixture
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -30,7 +30,7 @@ def pytest_configure(config):
     )
 
 
-class _TestAssets:
+class _DataEnv:
     """Test assets.
 
     Represents all the assets that have been added before the tests.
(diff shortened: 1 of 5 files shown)


jmorel:
@maeldebon did you mean to create this PR as draft or WIP ? The goal of the PR is very well defined but the code doesn't match.

tests/test_execution.py
```diff
@@ -101,15 +101,10 @@ def test_federated_learning_workflow(global_execution_env):
 
 
 @pytest.mark.slow
-def test_tuples_execution_on_different_nodes(global_execution_env):
+def test_tuples_execution_on_different_nodes(factory, client_1, client_2,
+                                             initial_assets_1, initial_assets_2):
     """Execution of a traintuple on node 1 and the following testtuple on node 2."""
     # add test data samples / dataset / objective on node 1
-    factory, initial_assets, network = global_execution_env
```
jmorel: The factory obtained through global_execution_env made sure assets created in the tests and initial_assets would share uuid, with this change they will have 2 different uuid. I'm not sure about the impacts though, this needs investigating.

maeldebon:
> @maeldebon did you mean to create this PR as draft or WIP ? The goal of the PR is very well defined but the code doesn't match.

Sorry, I could have created a draft instead. We are just trying this to possibly implement it later 🙂

tests/test_execution.py
samlesu: Yes, at first sight I don't think it would be a problem. But as you said, it requires further investigation.

tests/conftest.py
```diff
@@ -149,6 +149,67 @@ def global_execution_env():
         yield f, assets, n
 
 
+@pytest.fixture
```
samlesu: I think it would be good to execute this feature only once. It could be good to change its scope?

tests/conftest.py
```diff
@@ -149,6 +149,67 @@ def global_execution_env():
...
+def initial_assets():
```
samlesu: As this fixture will be used by almost all execution tests, I'm wondering if we could find a better name (shorter and single name). Any idea?

maeldebon: Maybe we could just name it `assets`?

tests/conftest.py
maeldebon: Yes indeed, I'll do that!

tests/conftest.py
samlesu: `assets` is already taken: it's a module name defining the responses from the backend.
So if we chose assets, we should IMO rename this module.

samlesu:
Thanks @maeldebon. I think it's a good idea to split the `global_execution_env` fixture into fixtures doing/returning one thing.

tests/conftest.py
samlesu: we could use `dataenv` ?

maeldebon: I was thinking of `data` but yes, it could be nice to use this one

tests/test_execution.py
```diff
@@ -10,14 +10,11 @@
 
 
 @pytest.mark.slow
-def test_tuples_execution_on_same_node(global_execution_env):
+def test_tuples_execution_on_same_node(factory, client, data_envs_1):
```
samlesu: I think it would be good to follow a unique convention when defining fixtures? From "more common" fixture to "less common":
network, factory, client(s), data_env(s) ?

tests/conftest.py
```diff
@@ -146,7 +195,19 @@ def global_execution_env():
             objectives.append(objective)
 
         assets = _TestAssets(datasets=datasets, objectives=objectives)
-        yield f, assets, n
+        return assets
+
+
+@pytest.fixture
+def data_envs_1(data_envs, client_1):
```
samlesu: Should be `data_env_1` (no plural)? Data environment for node 1.

tests/test_execution.py
```diff
@@ -55,21 +52,19 @@ def test_tuples_execution_on_same_node(global_execution_env):
 
 
 @pytest.mark.slow
-def test_federated_learning_workflow(global_execution_env):
+def test_federated_learning_workflow(factory, clients, client, data_envs):
```
samlesu: It seems strange to depend from clients and client.

What about initializing client from clients?
```python
client = clients[0]
```

tests/test_network.py
```diff
@@ -140,11 +140,11 @@ def test_add_composite_algo(factory, client):
     assert algo == algo_copy
 
 
-def test_list_nodes(network, client):
+def test_list_nodes(clients, client):
```
samlesu: Same here, use only the clients fixture

tests/conftest.py
```diff
@@ -100,11 +100,11 @@ def network():
     return _get_network()
 
 
-@pytest.fixture(scope='session')
-def global_execution_env():
-    """Network fixture with pre-existing assets in all nodes.
+@pytest.fixture(scope="session")
+def data_envs():
```
samlesu: @jmorel what do you think of this name? Any idea for a better name?

tests/test_execution.py
samlesu: @maeldebon what about this topic? Have you done some investigations?

tests/test_execution.py
```diff
@@ -10,14 +10,11 @@
...
+def test_tuples_execution_on_same_node(factory, client, data_envs_1):
     """Execution of a traintuple, a following testtuple and a following traintuple."""
-    factory, initial_assets, network = global_execution_env
-    client = network.clients[0]
 
-    initial_assets = initial_assets.filter_by(client.node_id)
-    dataset = initial_assets.datasets[0]
-    objective = initial_assets.objectives[0]
+    dataset = data_envs_1.datasets[0]
+    objective = data_envs_1.objectives[0]
```
samlesu: In most of the executions test we are instantiating these variables and using them only once.

Thanks to this PR, accessing the first dataset of the node 1 for instance is short and explicit.

We could avoid creating these variables and directly use `data_env_1.<asset_name>s[i]`. It could make the test shorter without decreasing its readability. What do you think @jmorel ?

tests/test_execution.py
```diff
@@ -55,21 +52,19 @@ def test_tuples_execution_on_same_node(global_execution_env):
...
     algo = client.add_algo(spec)
 
     # get first dataset of each client
     # because there is only one dataset created by client, we can get all the datasets
-    datasets = initial_assets.datasets
+    datasets = data_envs.datasets
 
     # check there is one dataset per node in the network
-    assert set([d.owner for d in datasets]) == set([c.node_id for c in network.clients])
+    assert set([d.owner for d in datasets]) == set([c.node_id for c in clients])
```
samlesu: We could even remove this line, the goal of the test is not to check the fixture.

tests/test_execution.py
maeldebon: That's what I tried to do. When you check the tests, you (almost) always need the `factory` fixture. The `network` is therefore used less than 10 times, because we directly import the profiles.

tests/conftest.py
maeldebon: Yes, I forgot to rename it, thank you!

tests/test_execution.py
maeldebon: Yes we could do this, or we could just rollback the changes by writing this:
```py
clients = network.clients[0]
```

Actually, we import `clients` because we need it a few lines below, in
```py
    assert set([d.owner for d in datasets]) == set([c.node_id for c in clients])
```

tests/test_execution.py
maeldebon: You're right, I'll do that too

tests/test_execution.py
maeldebon: Ok, I investigated this.
There is indeed a difference between the two UUIDs, but the root stays the same.

**In the `factory` fixture:**
```
##### f._uuid #####
 c306ac2a6e7c4cc087e74b66f46f5b51_test_tuples_execution_on_same_node
```

**In the `data_envs` fixture:**
```
##### f._uuid #####
 c306ac2a6e7c4cc087e74b66f46f5b51_global
```

tests/test_execution.py
maeldebon: It could be good for most of them, but I think we'll have to keep some variables for those that are more used 🙂

tests/conftest.py
jmorel: I'd rather have the same name for the class returned by the fixture (currently _TestAssets) and the fixture itself (currently data_envs).

`data_env` is pretty nice I think, it the class was renamed in `_DataEnv` and the fixture in `default_data_env` I'd be more than satisfied !

tests/test_execution.py
jmorel: Or we could have some more fixtures, `default_dataset` and `default_objective` that would return just these.

samlesu: Sounds like a good idea as each data_env has currently only a single dataset and a single objective.

tests/conftest.py
samlesu: Yes, I fully agree.

tests/conftest.py
```diff
@@ -145,8 +145,68 @@ def global_execution_env():
...
+@pytest.fixture
+def default_dataset(data_env_1):
+    """Fixture with pre-existing dataset in first node."""
+    return data_env_1.datasets[0]
+
+
+@pytest.fixture
+def default_objective(data_env_1):
+    """Fixture with pre-existing objective in first node."""
+    return data_env_1.objectives[0]
```
maeldebon: We could indeed use `default_dataset_1` & `default_objective_1` here, but I think that it is easier to read this way when we only need assets from one client.

samlesu: 👍

tests/test_execution.py
```diff
@@ -311,7 +260,8 @@ def test_aggregatetuple(global_execution_env):
 
 
 @pytest.mark.slow
-def test_aggregate_composite_traintuples(global_execution_env):
+def test_aggregate_composite_traintuples(factory, clients, default_datasets, default_objectives,
```
samlesu: minor: network should be in first or second position.

samlesu: I think it would be better to not mix all assets and assets from node.

So either use `default_datasets, default_objectives` or `default_dataset_1, default_dataset_2, default_objective_1, default_objective_2`.

maeldebon: I agree, I'll change this

maeldebon: Yes, it could indeed looks better

tests/test_execution.py
```diff
@@ -412,8 +356,7 @@ def test_aggregate_composite_traintuples(global_execution_env):
     # username/password are not available in the settings files.
 
     client = clients[0]
-    initial_assets = initial_assets.filter_by(client.node_id)
-    dataset = initial_assets.datasets[0]
+    dataset = default_datasets.sort_by(client.node_id)
```
samlesu: I don't think this is working as `default_datasets` is a list and a list instance has no method `sort_by`.
Could be replaced maybe with `dataset = default_datasets[0]`?

maeldebon: You're right, I'll do that!

tests/test_execution.py
```diff
@@ -55,27 +49,18 @@ def test_tuples_execution_on_same_node(global_execution_env):
...
-
-    # check there is one dataset per node in the network
-    assert set([d.owner for d in datasets]) == set([c.node_id for c in network.clients])
-
     # create 1 traintuple per dataset and chain them
     traintuple = None
     rank = 0
     compute_plan_id = None
-    for dataset in datasets:
+    for dataset in default_datasets:
```
jmorel: Could you add a coment explaining that default_datasets does contain datasets on each node and that as a result we can use it for federated learning.

tests/conftest.py
```diff
@@ -145,8 +145,68 @@ def global_execution_env():
...
+@pytest.fixture
+def data_env_2(default_data_env, client_2):
+    """Fixture with pre-existing assets in second node."""
+    return default_data_env.filter_by(client_2.node_id)
+
+
+@pytest.fixture
+def default_dataset(data_env_1):
+    """Fixture with pre-existing dataset in first node."""
+    return data_env_1.datasets[0]
```
jmorel: Since this is exactly the same as `default_dataset_1`, why not use the `default_dataset_1` fixture and return it instead of extracting it from data_env_1 ? This is pretty much the same thing but my thinking is that it'd make the chain of fixtures more explicit.

maeldebon: You're right!

PR #86, opened 2020-03-11
Add cleanModels option

Kelvin-M:
Companion PR : https://github.com/SubstraFoundation/substra-chaincode/pull/106 & https://github.com/SubstraFoundation/substra-backend/pull/172 & https://github.com/SubstraFoundation/substra/pull/135

### DIFF
--- a/substratest/factory.py
+++ b/substratest/factory.py
@@ -413,7 +413,7 @@ class _BaseComputePlanSpec(_Spec, abc.ABC):
 @dataclasses.dataclass
 class ComputePlanSpec(_BaseComputePlanSpec):
     tag: str
-    cleanModels: bool
+    clean_models: bool
 
 
 @dataclasses.dataclass
@@ -636,14 +636,14 @@ class AssetsFactory:
             tag=tag,
         )
 
-    def create_compute_plan(self, tag='', cleanModels=False):
+    def create_compute_plan(self, tag='', clean_models=False):
         return ComputePlanSpec(
             traintuples=[],
             composite_traintuples=[],
             aggregatetuples=[],
             testtuples=[],
             tag=tag,
-            cleanModels=cleanModels
+            clean_models=clean_models
         )
 
     def update_compute_plan(self, compute_plan):
--- a/substratest/factory.py
+++ b/substratest/factory.py
@@ -413,6 +413,7 @@ class _BaseComputePlanSpec(_Spec, abc.ABC):
 @dataclasses.dataclass
 class ComputePlanSpec(_BaseComputePlanSpec):
     tag: str
+    cleanModels: bool
 
 
 @dataclasses.dataclass
@@ -635,13 +636,14 @@ class AssetsFactory:
             tag=tag,
         )
 
-    def create_compute_plan(self, tag=''):
+    def create_compute_plan(self, tag='', cleanModels=False):
         return ComputePlanSpec(
             traintuples=[],
             composite_traintuples=[],
             aggregatetuples=[],
             testtuples=[],
             tag=tag,
+            cleanModels=cleanModels
         )
 
     def update_compute_plan(self, compute_plan):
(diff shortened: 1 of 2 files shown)


substratest/factory.py
```diff
@@ -413,6 +413,7 @@ def add_testtuple(self, objective, traintuple_spec, tag=None):
 @dataclasses.dataclass
 class ComputePlanSpec(_BaseComputePlanSpec):
     tag: str
+    cleanModels: bool
```
```suggestion
    clean_models: bool
```
Cf my comment in the backend repository, substra backend should accept snake case fields to be consistent with the other endpoints.

tests/test_execution_compute_plan.py
```diff
@@ -332,7 +332,7 @@ def test_compute_plan_aggregate_composite_traintuples(global_execution_env):
     previous_composite_traintuple_specs = []
 
     cp_spec = factory.create_compute_plan(
-        cleanModels=True
```
samlesu: Maybe we could add a dedicated test for this feature. 
IMO, it's not straightforward to understand what's the purpose of setting this option to true and how to test it. 

We could maybe implement the following scenario:

1. create a small CP with clean_models enabled
2. wait for the CP to be executed
3. create a new tuple based on an intermediary model
4. check the tuple failed

If we don't want to create a new test, we could just add steps 3 and 4 to this test.

jmorel: Maybe we could also check the `shared_path` if available to make sure objects were removed physically.

tests/test_execution_compute_plan.py
```diff
@@ -379,6 +379,68 @@ def test_compute_plan_aggregate_composite_traintuples(global_execution_env):
         assert t.status == assets.Status.done
 
 
+@pytest.mark.slow
+def test_compute_plan_remove_intermediary_model(global_execution_env):
+    """
+    Create a simple compute plan with clean_models at true, see it done and
+    create a test tuple on a intermediary model. Expect it to fail at the
+    execution.
```
samlesu: The test is creating a traintuple and not a testtuple.

tests/test_execution_compute_plan.py
```diff
@@ -379,6 +379,68 @@ def test_compute_plan_aggregate_composite_traintuples(global_execution_env):
...
+    cp_spec.add_testtuple(
+        objective=objective,
+        traintuple_spec=traintuple_spec_2,
+    )
+
+    cp = client.add_compute_plan(cp_spec).future().wait()
+
+    # All the train/test tuples should succeed
+    for t in cp.list_traintuple() + cp.list_testtuple():
+        assert t.status == assets.Status.done
```
samlesu: This is not necessary as it is already checked in the wait() method.

tests/test_execution_compute_plan.py
Kelvin-M: What do you mean by the `shared_path` ?

jmorel: For the local test environment (setup through skaffold), there is a `shared_path` pointing to the servermedias folder (see `local_values.json`) that is accessible both by the tests and the platform. This was done to test adding data samples using a local path instead of an upload.

My thinking was that this path probably also contains the generated models and that we could check if physically the models were deleted or not.

tests/test_execution_compute_plan.py
thibaultrobert: Doesn't it just test that it finished but not done?

samlesu: `wait()` raises by default an exception if the status is canceled or failed. It can be changed if setting the input argument `raises` to `False`.

tests/test_execution_compute_plan.py
Kelvin-M: Checking model existence can be an issue because we cannot be sure that we will have an access to the models volume in the future !

PR #88, opened 2020-03-13
Nightly tests

Kelvin-M:
This PR introduces a CI script to run all the tests from `substra-tests` against a fresh substra deployment (in GKE).

The script is set up in Travis to [run every 24h](https://travis-ci.org/github/SubstraFoundation/substra-tests/settings). See a sample run [here](https://travis-ci.org/github/SubstraFoundation/substra-tests/builds/673169541).

This PR also sets up slack/email notification. We also built a [Google Cloud Function](https://console.cloud.google.com/functions/list?hl=fr&project=substra-208412) which deletes stale clusters after 24h (in case a Travis build gets killed half way through and the cluster is still up, or someone at home testing the script commented the cluster deletion part). It is run [using a Google Cloud Scheduler](https://console.cloud.google.com/cloudscheduler?hl=fr&project=substra-208412&folder=&organizationId=). See notification examples in #test-notifs.

## What `run-ci.py` does

- Create a new GKE cluster
- Clone the repos (hlf-k8s, substra-backend, substra-tests)
- Build the docker images in GKE (with docker caching, 1 week ttl by default)
- Deploy with `skaffold deploy`
- Wait for the substra application stack to be up and ready (i.e. wait for the substra-backend servers readiness probe)
- Run the full 'substra-tests' test suite
- Destroy the cluster
  
## Retries

The substra deployment isn't fully deterministic (1 time out of 10, the pods never get in the ready state). Fixing this issue is out of the scope of this PR. 

However, we want the results of the nightly Travis CI runs to be reliable. In particular, we want to notify devs of CI runs failures only if the failure happens *consistently*. To achieve that goal, we retry the CI script up to 3 times before definitely marking the build as "failed". 

## Run time

- with warm cache: ~14min
- with cold cache: ~24min
- worst case scenario (cold cache + 2 retries): 55 min (Travis will auto-kill the builds after 50 min and mark it as 'errored')

## Release checklist

- [x] Point Travis slack notifications to the right channel
- [x] Point Google Scheduler slack notifications to the right channel
- [x] Merge [this PR](https://github.com/SubstraFoundation/substra-tests/pull/98)


scripts/deploy.template.sh
```diff
@@ -20,7 +20,8 @@ helm init --upgrade --force-upgrade --wait
 curl -Lo skaffold https://storage.googleapis.com/skaffold/releases/v1.0.1/skaffold-linux-amd64 && chmod +x skaffold && mv skaffold /usr/local/bin
 
 # Clone repos
-mkdir src
+rm -rf src
```
AurelienGasser: why rm if `mkdir -p` next ?

.travis.yml
```diff
@@ -0,0 +1,35 @@
...
+    - master
+    - nightly-tests
+
+before_install:
+  # Decrypt secrets
+  - openssl aes-256-cbc -K $encrypted_36785d287aa1_key -iv $encrypted_36785d287aa1_iv
+    -in ci/keys/substra-208412-3be0df12d87a.json.enc -out ci/keys/substra-208412-3be0df12d87a.json
+    -d
+  - openssl aes-256-cbc -K $encrypted_3657ea9362f7_key -iv $encrypted_3657ea9362f7_iv
+    -in ci/keys/kaniko-secret.json.enc -out ci/keys/kaniko-secret.json -d
```
Kelvin-M: Can you provide/add link to the travis documentation for this step :) ?

.travis.yml
```diff
@@ -0,0 +1,35 @@
...
+  # Install helm (v2.14.2)
+  - curl https://get.helm.sh/helm-v2.14.2-linux-amd64.tar.gz -o helm-v2.14.2-linux-amd64.tar.gz
+  - tar xzf helm-v2.14.2-linux-amd64.tar.gz
+  - mv linux-amd64/helm linux-amd64/tiller $HOME/bin/
+  # Install kubectl
+  - curl -LO https://storage.googleapis.com/kubernetes-release/release/`curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt`/bin/linux/amd64/kubectl
+  - chmod +x ./kubectl
+  - mv ./kubectl $HOME/bin/
```
Kelvin-M: @ClementGautier Is there a way to install package in travis or maybe a image exist to avoid installing helm and kubectl manually :) ?

.travis.yml
```diff
@@ -0,0 +1,35 @@
...
+  - CLUSTER_NAME="substra-tests-$(date -u +'%Y-%m-%d-%Hh%M')-$(cat /dev/urandom | tr -dc 'a-z0-9' | fold -w 8 | head -n 1)"
+  - CLUSTER_ZONE="europe-west4-a"
+  - cd ci/
+  - ./run-tests.sh --keys-directory=./keys/ --cluster-name=${CLUSTER_NAME} --cluster-zone=${CLUSTER_ZONE}
+
+after_script:
+  # run-tests.sh should delete the cluster, but make sure the cluster really gets deleted
+  # in case travis kills the script.
+  # Todo: I'm not sure this actually is always called (script timeout, build canceled, etc)
+  # Todo: this doesn't work at all: we're not logged into google at this point
```
Kelvin-M: We should add a login step here :)
If we see travis job livecycle, after-script is executed :)

.travis.yml
Kelvin-M: It's add addons :) 
https://docs.travis-ci.com/user/job-lifecycle/#the-job-lifecycle

.travis.yml
AurelienGasser: +1 adding

.travis.yml
AurelienGasser: Lies !!! :p https://travis-ci.org/github/SubstraFoundation/substra-tests/builds/667720069

AurelienGasser: on cancel: https://travis-ci.org/github/SubstraFoundation/substra-tests/builds/667720069

AurelienGasser: Update: we decided to use a GCP cron job to delete clusters in case of uncaught errors

.travis.yml
Kelvin-M: If I understand well, travis `addon` is just a syntax, it will execute nearly the same command (apt, snap, brew, etc.)
It's just a travis way to install dependencies. Nevertheless, I don't know if it's better or not to use it ! 
@ClementGautier  Do you have some insights to share on it ?

ci/run-ci.sh
```diff
@@ -0,0 +1,135 @@
...
+#        * Clone the repos:
+#           ~ hlf-k8s
+#           ~ substa-backend
+#           ~ substra-tests
+#           ~ substra
+#        * Build the images to the private registry (kaniko)
+#        * Deploy (helm)
+#   - Wait for the substra application stack to be up and ready (i.e. wait for the substra-backend servers readiness probe)
+#   - Run the full 'substra-tests' test suite
+#   - Destroy the cluster
```
samlesu, line 19: :100:

skaffold.yaml
```diff
@@ -0,0 +1,36 @@
+# Copyright 2018 Owkin, inc.
```
samlesu, line 1: Not sure if it's relevant, but could be moved to the `ci/` folder (it would make sense if this yaml file is only used for the ci) ?
Same question for the charts and docker folders?

.travis.yml
```diff
@@ -0,0 +1,54 @@
+language: minimal
+
+branches:
+  only:
+    - master
+    - nightly-tests
```
samlesu, line 6: Should we remove the nightly-tests branch before merging?

Kelvin-M: Yes, as all references to nightly tests too !

Kelvin-M:
> Thanks you guys, it looks awesome ! Can't wait to have these notifications!
> I'm not the most experienced to review this PR. Maybe we could ask @AlexandrePicosson to have a look?
> 
> A couple of question:
> 
> * As far as I understand a notifications is sent to Slack and by emails. Could you add me or tell me where I can see the slack ones?
> * In case of failure do we have the logs or a link to the report?

The notifications parameters are in the `travis.yaml` file. We should decide of these parameters in a meeting (who/where/what notification to send)

In all case, a link to the travis dashbord result is sent :)

jmorel:
I tried running very naively `run-ci.sh` with this result:
```sh
./run-ci.sh
KEYS_DIR     = /Users/jeremy/.local/
CLUSTER_NAME = substra-tests

# Always delete the cluster, even if the script fails
delete-cluster() {
    yes | gcloud container clusters delete ${CLUSTER_NAME} --zone ${CLUSTER_ZONE} --project ${CLUSTER_PROJECT}
}
./run-ci.sh: line 73: `delete-cluster': not a valid identifier
```

Kelvin-M:
> I tried running very naively `run-ci.sh` with this result:
> 
> ```shell
> ./run-ci.sh
> KEYS_DIR     = /Users/jeremy/.local/
> CLUSTER_NAME = substra-tests
> 
> # Always delete the cluster, even if the script fails
> delete-cluster() {
>     yes | gcloud container clusters delete ${CLUSTER_NAME} --zone ${CLUSTER_ZONE} --project ${CLUSTER_PROJECT}
> }
> ./run-ci.sh: line 73: `delete-cluster': not a valid identifier
> ```

Looks like a `shell` issue https://stackoverflow.com/questions/53364885/mac-osx-terminal-not-a-valid-identifier-on-function-definition

jmorel:
Thanks for the fix !

I'm now running into another issue:
```
# Deploy
# TODO: change git.substraTests.branch to master (necessary for now because skaffold.yaml doesn't exist in master)`
helm install ${CHARTS_DIR}/substra-tests-stack \
    --kube-context ${KUBE_CONTEXT} \
    --name substra-tests-stack \
    --set image.repository=${IMAGE_SUBSTRA_TESTS_DEPLOY_REPO} \
    --set image.tag=${IMAGE_SUBSTRA_TESTS_DEPLOY_TAG} \
    --set deploy.defaultRepo=${REGISTRY} \
    --set git.substraTests.branch=nightly-tests \
    --wait
+ helm install /Users/jeremy/substra/substra-tests/ci/../charts/substra-tests-stack --kube-context gke_substra-208412_europe-west4-a_substra-tests --name substra-tests-stack --set image.repository=substrafoundation/substra-tests-stack --set image.tag=latest --set deploy.defaultRepo=10.56.0.14:5000 --set git.substraTests.branch=nightly-tests --wait
Error: transport is closing
delete-cluster
+ delete-cluster
+ yes
+ gcloud container clusters delete substra-tests --zone europe-west4-a --project substra-208412
The following clusters will be deleted.
 - [substra-tests] in [europe-west4-a]

Do you want to continue (Y/n)?
Deleting cluster substra-tests...
............................................
```

skaffold.yaml
AurelienGasser: - the docker/chart/skaffold for `substra-test` can potentially be used for other contexts (not CI)
- the docker/chart for `substra-test=stack` is designed for the CI only

I would leave the structure as it is now for simplicity and consistenct with other repos, although I don't have super stromg feelings one way or another.

samlesu: I don't have a strong opinion on this also, so as you said let's keep this as it is. Thanks for the answer.

AurelienGasser:
@jmorel looks like it could have been a transient connectivity issue on your end. Mind trying again?

AurelienGasser:
@jmorel also please change the value of CLUSTER_NAME to something like "substra-tests-jeremy" before running. It's not very clear in the comments, I'll fix that.

you can also use `./run-ci.sh --cluster-name=substra-tests-jeremy`

jmorel:
With your command @AurelienGasser it ran fine. And I'm pleased to report that

```
============= 77 passed, 6 skipped, 1 xfailed in 271.40s (0:04:31) =============
```

Well done!

docker/substra-tests-stack/bin/deploy.sh
```diff
@@ -0,0 +1,39 @@
...
+cd src
+git clone --depth 1 https://github.com/SubstraFoundation/substra-backend.git
+git clone --depth 1 https://github.com/SubstraFoundation/hlf-k8s.git
+git clone --depth 1 https://github.com/SubstraFoundation/substra-tests.git --branch "${SUBSTRA_TESTS_BRANCH}"
+cd ..
+
+# Patch skaffold files for kaniko
+python -m venv .venv
+source .venv/bin/activate
+pip install pyyaml
```
AlexandrePicosson, line 24: Why are you running the pip install command at runtime instead of running it in your Dockerfile ?
You could use something like this to install directly in your virtual environment: https://pythonspeed.com/articles/activate-virtualenv-dockerfile/

Kelvin-M: We can install pyyaml directly within the Dockerfile. No need to use a virtualenv in docker.
It's surely a remaining from previous script setup

charts/substra-tests-stack/templates/rbac.yaml
```diff
@@ -40,5 +28,17 @@ subjects:
...
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: tiller-user
+  namespace: kube-system
+rules:
+rules:
+- apiGroups: ["", "batch", "extensions", "apps"]
+  resources: ["*"]
+  verbs: ["*"]
```
AlexandrePicosson, line 44: Helm really needs these access to run ? In the docs they just reference `pods/portforward` and `pods`.
This is the role your pod runnning helm will use, not the role for the tiller pod.
Your tiller pod should already be created and cluster-admin according to this line : https://github.com/SubstraFoundation/substra-tests/blob/553994e918e760568068c313b8d478b803e72be2/ci/run-ci.sh#L138 .

docker/substra-tests-stack/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+# Install helm (v2.14.2)
+RUN curl https://get.helm.sh/helm-v2.14.2-linux-amd64.tar.gz -o helm-v2.14.2-linux-amd64.tar.gz
+RUN tar xzf helm-v2.14.2-linux-amd64.tar.gz
+RUN mv linux-amd64/helm linux-amd64/tiller /usr/local/bin/
+
+# Install skaffold (v1.0.1)
+RUN curl -Lo skaffold https://storage.googleapis.com/skaffold/releases/v1.0.1/skaffold-linux-amd64 && chmod +x skaffold && mv skaffold /usr/local/bin
+
+# Install pyyaml
+pip3 install pyyaml
```
AlexandrePicosson, line 19: This should probably be `RUN pip3 install pyyaml`.

Kelvin-M: Yes my bad ^^

charts/substra-tests-stack/templates/rbac.yaml
Kelvin-M: I've tested with the documentation access, but it does not work. I have to spend more time to see the real apiGroups, resources and verbs needed.

Kelvin-M: It seems that the documentation show miniaml rbac for helm to interact with tiller but only interact ^^

AlexandrePicosson: But helm should only interact with tiller as tiller is the one in charge of creating ressources in the k8s cluster.
I think the issue is related to the fact that we have the --upgrade flag on the helm init command here: https://github.com/SubstraFoundation/substra-tests/blob/143cc4a76b80e4fad9d036808bfebf120f66cf1f/docker/substra-tests-stack/bin/deploy.sh#L10 So helm needs the rights to deploy a new tiller.

But I think the upgrade is mandatory as you can't be certain of the helm version the user used to run the ci/run-ci.sh script.

Kelvin-M: Moreover, we heavily use skaffold which need to operate on pods, etc. for the build and the run part.

Kelvin-M: @AlexandrePicosson  I've spend some time on it and updated https://github.com/SubstraFoundation/substra-tests/pull/88/commits/f7282e8b9903f782dc76435aafdc9a38211107fc

If you have some ideas feel free to contribute with a commit :) !

AurelienGasser:
@AlexandrePicosson @samlesu @jmorel This is ready for another review. (cc @Kelvin-M @thibaultrobert @ClementGautier)

The PR description has been updated.

Note that the code still has mentions of the "nightly-tests" branch: this is intentional to allow running test builds locally and in travis. Those will be replaced with "master" as the last step before merging.

.travis.yml
ClementGautier: Suing addons may be more robust as it uses a miror on travis side to cache packages. It's also easier to read in my opinion but that's a matter of taste.

AurelienGasser: (I should have mentioned it before) I initially used addons but snap only has helm 2.15 (not 2.14). I couldn't get things to work with 2.15. We could give it another shot though.

AurelienGasser:
@ClementGautier The decryption key is stored as a secure token. So only a Travis build can decrypt the SA key. The attack vectors are the same as if the SA key was a secret token directly. Does that make sense?

ClementGautier:
> @ClementGautier The decryption key is stored as a secure token. So only a Travis build can decrypt the SA key. The attack vectors are the same as if the SA key was a secret token directly. Does that make sense?

Yup, totally missed that, it's fine then :)

AurelienGasser:
PR to merge right before the merge to master: https://github.com/SubstraFoundation/substra-tests/pull/98

ci/run-ci.py
```diff
@@ -0,0 +1,475 @@
...
+        build_images(configs)
+        wait_for_cluster()
+        get_kube_context()
+        setup_tiller()
+        deploy_all(configs)
+        is_success = run_tests()
+
+    except Exception as e:
+        print(f'FATAL: {e}')
+        is_success=False
```
samlesu, line 464: Why printing the exception?

It would be simpler to not catch it: the traceback will be printed in stdout and the result code will be 1.

.travis.yml
```diff
@@ -0,0 +1,58 @@
+language: python
+python:
+  - "3.7"
+
+branches:
+  only:
```
inalgnu, line 6: maybe we should add a regular expression branch, something like `/^tobetested-.*$/` To be able to run tests using a pattern

ci/run-ci.py
AurelienGasser: I added a `print` here is so that the exception is displayed _before_ the `finally` block is executed, not after. If it's displayed after, the logs can be hard to read. I also suspect that if we abort the script during `finally` (Ctrl+C), the exception might not be displayed at all?

.travis.yml
AurelienGasser: Good idea, but I don't think that in practice people will remember to name their branch correctly. Also, in new branches, devs can override the `branches` property to allow Travis builds for their branch.

ci/run-ci.py
```diff
@@ -0,0 +1,475 @@
...
+        f'SUBSTRA_TESTS_BRANCH\t= {SUBSTRA_TESTS_BRANCH}\n'
+        f'SUBSTRA_BACKEND_BRANCH\t= {SUBSTRA_BACKEND_BRANCH}\n'
+        f'HLF_K8S_BRANCH\t\t= {HLF_K8S_BRANCH}\n'
+        f'KANIKO_CACHE_TTL\t= {KANIKO_CACHE_TTL}\n'
+    )
+
+
+def gcloud_login():
+    print('# Log into Google Cloud')
+    call(f'gcloud auth activate-service-account {SERVICE_ACCOUNT} --key-file={KEYS_DIR}/{KEY_SERVICE_ACCOUNT}')
```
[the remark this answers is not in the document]
AurelienGasser: I looked into the `gcloud builds` python package, and it just got out of alpha. The doc is really not great so I didn't see much value in using it (yet, at least). For the other parts (`gcloud cluster`, `git`, `kubernetes`) we could explore it. @Kelvin-M thoughts?

ci/run-ci.py
```diff
@@ -0,0 +1,475 @@
...
+
+
+def arg_parse():
+
+    global KEYS_DIR
+    global CLUSTER_NAME
+    global SUBSTRA_TESTS_BRANCH
+    global SUBSTRA_BACKEND_BRANCH
+    global HLF_K8S_BRANCH
+    global KANIKO_CACHE_TTL
```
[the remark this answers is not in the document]
AurelienGasser: I agree in principle, but wouldn't the constraints introduced by doing that outweight the benefits?

charts/substra-tests/Chart.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
+appVersion: "1.0"
+description: A Helm chart for building and running substra tests
```
inalgnu, line 3:
```suggestion
description: A Helm chart for building and running Substra tests
```

ci/.gitignore
```diff
@@ -0,0 +1 @@
+src/
```
inalgnu, line 1: missing EOF

ci/run-ci.py
```diff
@@ -0,0 +1,475 @@
...
+    return subprocess.check_call([cmd], shell=True)
+
+
+def call_output(cmd, print_cmd=True):
+    if print_cmd:
+        print(f'+ {cmd}')
+    return subprocess.check_output([cmd], shell=True, stderr=subprocess.STDOUT).decode().strip()
+
+
+def cluster_name(value):
```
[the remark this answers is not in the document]
AurelienGasser: It is used! In `arg_parse()`

```
parser.add_argument('-N', '--cluster-name', type=cluster_name, [...]
```

ci/cloudbuild/substra-backend.yaml
```diff
@@ -0,0 +1,18 @@
...
+  - --cache=true
+  - --dockerfile=substra-backend/docker/substra-backend/Dockerfile
+  - --context=substra-backend
+  - --cache-ttl=${_KANIKO_CACHE_TTL}
+tags:
+- ${_BUILD_TAG}
+substitutions:
+    _BUILD_TAG: "substra-tests"
+    _COMMIT: "HEAD"
+    _BRANCH: "master"
```
inalgnu, line 16: just trying to understand, why we don't use ${_BRANCH} here?

ci/.gitignore
AurelienGasser: fixed: https://github.com/SubstraFoundation/substra-tests/pull/88/commits/9ea1e9c64cee8aeecbd9378bd37131c82889810f

ci/run-ci.py
```diff
@@ -0,0 +1,475 @@
...
+    try:
+        call(f'kubectl --context {KUBE_CONTEXT} exec {substra_tests_pod} -n substra-tests -- make test')
+        return True
+    except subprocess.CalledProcessError as e:
+        print('FATAL: `make test` completed with a non-zero exit code. Did some test(s) fail?')
+        return False
+
+
+def main():
+    is_success=False
```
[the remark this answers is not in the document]
AurelienGasser: fixed: https://github.com/SubstraFoundation/substra-tests/pull/88/commits/9ea1e9c64cee8aeecbd9378bd37131c82889810f

ci/run-ci.py
```diff
@@ -0,0 +1,475 @@
...
+#   - Stale cluster deletion script:
+#       https://console.cloud.google.com/functions/details/us-central1/clean-substra-tests-ci-deployment
+import os
+import time
+import json
+import shutil
+import string
+import random
+import argparse
+import functools
```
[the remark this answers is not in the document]
AurelienGasser: fixed: https://github.com/SubstraFoundation/substra-tests/pull/88/commits/9ea1e9c64cee8aeecbd9378bd37131c82889810f

charts/substra-tests/Chart.yaml
AurelienGasser: fixed: https://github.com/SubstraFoundation/substra-tests/pull/88/commits/9ea1e9c64cee8aeecbd9378bd37131c82889810f + https://github.com/SubstraFoundation/substra-tests/pull/88/commits/4822f30b6803416975a48eb2a9053690b44af2ab

ci/cloudbuild/substra-backend.yaml
AurelienGasser: The `substitutions` section just defines the variables of our script. This section is mandatory, and the values you see here are the (mandatory) default values.

inalgnu: thank's @AurelienGasser

.travis.yml
inalgnu: Ok thank you for the details @AurelienGasser

.travis.yml
```diff
@@ -0,0 +1,58 @@
...
+      # - change target channel in the slack UI
+      # - regenerate encrypted payload below (https://docs.travis-ci.com/user/notifications/#configuring-slack-notifications)
+      - secure: UC3111xkSw2nluYtQb4Uqpufkc51WeY4pbm3TJdKJ5ydcTUrvDpcYHic8/5xfBLgU2EkBN6795bu9CNSZHDqO43QkMVSfz3bBea3S3r/bTDng33FaWUMaU1qLI2YnLrlO9STEL0bFGWc0OiC0BPW8kJvKYkk6szukB8OdvOt0mqApNcTRGkiofAlewk1ot1m1vTmMMryZF0cyTWyYz10dxacGSOIvILJECNP+JqpnjRzCat5sfjUjVS5jESBAIw++I3sLLSp7QTgDjY2SpIeEh0ksd6rVZUx2I1FAG3XbhlCAqWeqp6gG6CJbzl6q9Dm9cKgUWPlw5zBQ5mwl+Qqncui/Nt5BSa+F/WFxpYKKgoIIhBJvobpsZ96U0pIMpr5hZSPjs2axRPSSFjt8aiEp4fqojFj+50AXgp0qip0dZoG41n4VVCtaHHxwZUVhJ48Hkqv2gWH4pT56oYoZylT+E9fKAuFhDm4f0WeQdXo9pB1cqEM1xSHMIX8M/YiXxNDZE073MjReXVsArhqHQwoGBIP6ITYRPwUjqargnuUY6A/SoelV4e+mU2iyLjyaXNmx+LvQL+wnrP9yDXI2KsovFLZpetpm7arPramu/jAJE+d+wxqUPnq22TOOrtoZCGZbFn98D+fbjCzEw+b/slovdsMjCDcYAoYp7HttqA0fMM=
+    on_success: change
+    on_failure: always
+    template:
+      - "Nightly test <%{build_url}|#%{build_number}> of %{repository_name}@%{branch} %{result} in %{duration}"
+    if: type = cron AND (branch = master OR branch = nightly-tests) AND fork = false
+
+  email:
```
inalgnu, line 51: could be pleasant to have a subject pattern here to be able to filter nightly-tests emails

ci/run-ci.py
```diff
@@ -0,0 +1,474 @@
...
+CLUSTER_MACHINE_TYPE = 'n1-standard-8'
+
+CLUSTER_VERSION = '1.15.11-gke.1'
+CLUSTER_PROJECT = 'substra-208412'
+CLUSTER_ZONE = 'europe-west4-a'
+
+SERVICE_ACCOUNT = 'substra-tests@substra-208412.iam.gserviceaccount.com'
+KEY_SERVICE_ACCOUNT = 'substra-208412-3be0df12d87a.json'
+
+SUBSTRA_TESTS_BRANCH = 'nightly-tests'
```
inalgnu, line 57: @todo change this :-)

ci/run-ci.py
```diff
@@ -0,0 +1,474 @@
...
+
+    KEYS_DIR = args['keys_directory']
+    SUBSTRA_TESTS_BRANCH = args['substra_tests']
+    SUBSTRA_BACKEND_BRANCH = args['substra_backend']
+    HLF_K8S_BRANCH = args['hlf_k8s']
+    if args['no_cache']:
+        KANIKO_CACHE_TTL = '-1h'
+
+    print(
+        f'💃💃💃\n'
```
inalgnu, line 141: 😁

ci/run-ci.py
AurelienGasser: It's here: https://github.com/SubstraFoundation/substra-tests/pull/98

It's not merged yet because we want to be able to keep testing changes in Travis if necessary. If we reference the master branch here directly  then we can't run the tests in travis because the master branch is still the old version.

ci/run-ci.py
```diff
@@ -0,0 +1,474 @@
...
+         'images': ['substra-tests'],
+         'commit': commit_tests,
+         'branch': SUBSTRA_TESTS_BRANCH},
+    ]
+
+
+def clone_repository(dirname, url, branch, commit=None):
+    call(f'git clone -q --depth 1 {url} --branch "{branch}" {dirname}')
+
+    if commit is None:
```
inalgnu, line 255: it's a bit confusing for me here if the commit is given do we checkout the commit? or just return it? what is the purpose of passing a commit to this function?

AurelienGasser: I think this was done to eventually allow specifying a commit hash at the top level of the script, but it wasn't finished. @Kelvin-M would know more

ci/run-ci.py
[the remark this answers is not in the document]
Kelvin-M: Yes, same as Auélien.
We could use only a mutable structure as dictionnary  and pass it as reference in each call.
But we should remember that's a script to launch the test and it's easier to have something simple.

ci/run-ci.py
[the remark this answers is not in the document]
Kelvin-M: We should remember that's a script to launch the test and it's easier to have something simple. Using sdk could be a solution, but we will increase dependencies to external libraries (often not as good as cli). Here we have something which use only the python standard library.

ci/run-ci.py
Kelvin-M: Yes

ci/run-ci.py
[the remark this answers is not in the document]
samlesu: I do agree, it's debatable. In my opinion, it would make the code simpler to use a mutable structure.

As said by @Kelvin-M and as said in my main comment, it's a script to launch the tests, so it's good for me.

ci/run-ci.py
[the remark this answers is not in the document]
samlesu: Thanks! I haven't seen that.

ci/run-ci.py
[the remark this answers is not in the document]
samlesu: Ok, thanks.

ci/run-ci.py
samlesu: Sounds good to me.

Kelvin-M:
@inalgnu @samlesu @jmorel 
We surely need a syncro when we want to merge it into master :
We will have to merge the following PR into this one before :)
https://github.com/SubstraFoundation/substra-tests/pull/98

.travis.yml
AurelienGasser: Looks like this isn't supported unfortunately: https://config.travis-ci.com/ref/notifications/email

PR #90, opened 2020-03-20
Add user metadata

maeldebon:
🔗 **Related PRs:**
- [substra](https://github.com/SubstraFoundation/substra/pull/143)
- [substra-chaincode](https://github.com/SubstraFoundation/substra-chaincode/pull/111)
- [substra-backend](https://github.com/SubstraFoundation/substra-backend/pull/186)
- [substra-frontend (coming soon)](#)

The goal of these PRs is to add a `metadata` field in every asset in order to be able to store information related to this asset. It could be, for instance, a creation_ts, an epoch, a project_name, ...

Update 04/24:
> The following assets are missing in the tests:
> - dataset,
> - aggregatetuple (WIP),
> - compute_plan.

> The issue I currently have is that the metadata field returned by the aggregatetuple tests is set to an empty dict, but I pass:
{"foo": "bar"}
> I need to find a way to get the right output.

Update 06/03:
> The compute_plan asset cannot have a metadata field for now. It could be done later.

### DIFF
--- a/substratest/assets.py
+++ b/substratest/assets.py
@@ -201,7 +201,7 @@ class Dataset(_Asset):
     name: str
     owner: str
     objective_key: str
-    metadata: typing.Dict[str, str] = None
+    metadata: typing.Dict[str, str]
     permissions: Permissions
     # the JSON data returned by list_dataset doesn't include the following keys at all
     # they are only included in the result of get_dataset
@@ -337,7 +337,7 @@ class Testtuple(_Asset, _FutureMixin):
     certified: bool
     rank: int
     tag: str
-    metadata: typing.Dict[str, str] = None
+    metadata: typing.Dict[str, str]
     log: str
 
 
@@ -350,7 +350,7 @@ class ComputePlan(_Asset, _ComputePlanFutureMixin):
     testtuple_keys: typing.List[str]
     id_to_key: typing.Dict[str, str]
     tag: str
-    metadata: typing.Dict[str, str] = None
+    metadata: typing.Dict[str, str]
 
     def __init__(self, *args, **kwargs):
         kwargs['traintuple_keys'] = kwargs.get('traintuple_keys') or []
--- a/substratest/assets.py
+++ b/substratest/assets.py
@@ -350,7 +350,6 @@ class ComputePlan(_Asset, _ComputePlanFutureMixin):
     testtuple_keys: typing.List[str]
     id_to_key: typing.Dict[str, str]
     tag: str
-    metadata: typing.Dict[str, str]
 
     def __init__(self, *args, **kwargs):
         kwargs['traintuple_keys'] = kwargs.get('traintuple_keys') or []
--- a/substratest/assets.py
+++ b/substratest/assets.py
@@ -299,7 +299,7 @@ class Aggregatetuple(_Asset, _FutureMixin):
     tag: str
     metadata: typing.Dict[str, str]
     log: str
-    in_models: typing.List[InModel]
+    in_models: typing.List[InModel] = None
     out_model: typing.Optional[OutModel]
 
 
--- a/substratest/assets.py
+++ b/substratest/assets.py
@@ -201,6 +201,7 @@ class Dataset(_Asset):
     name: str
     owner: str
     objective_key: str
+    metadata: typing.Dict[str, str] = None
     permissions: Permissions
     # the JSON data returned by list_dataset doesn't include the following keys at all
     # they are only included in the result of get_dataset
@@ -212,6 +213,7 @@ class _Algo(_Asset):
     key: str
     name: str
     owner: str
+    metadata: typing.Dict[str, str]
     permissions: Permissions
 
 
@@ -231,6 +233,7 @@ class Objective(_Asset):
     key: str
     name: str
     owner: str
+    metadata: typing.Dict[str, str]
     permissions: Permissions
     test_dataset: typing.Optional[ObjectiveDataset]
 
@@ -279,6 +282,7 @@ class Traintuple(_Asset, _FutureMixin):
     compute_plan_id: str
     rank: int
     tag: str
+    metadata: typing.Dict[str, str]
     log: str
     in_models: typing.Optional[typing.List[InModel]]
     out_model: typing.Optional[OutModel]
@@ -293,6 +297,7 @@ class Aggregatetuple(_Asset, _FutureMixin):
     compute_plan_id: str
     rank: int
     tag: str
+    metadata: typing.Dict[str, str]
     log: str
     in_models: typing.List[InModel]
     out_model: typing.Optional[OutModel]
@@ -316,6 +321,7 @@ class CompositeTraintuple(_Asset, _FutureMixin):
     compute_plan_id: str
     rank: int
     tag: str
+    metadata: typing.Dict[str, str]
     log: str
     in_head_model: typing.Optional[InModel]
     in_trunk_model: typing.Optional[InModel]
@@ -331,6 +337,7 @@ class Testtuple(_Asset, _FutureMixin):
     certified: bool
     rank: int
     tag: str
+    metadata: typing.Dict[str, str] = None
     log: str
 
 
@@ -343,6 +350,7 @@ class ComputePlan(_Asset, _ComputePlanFutureMixin):
     testtuple_keys: typing.List[str]
     id_to_key: typing.Dict[str, str]
     tag: str
+    metadata: typing.Dict[str, str] = None
 
     def __init__(self, *args, **kwargs):
         kwargs['traintuple_keys'] = kwargs.get('traintuple_keys') or []
--- a/substratest/factory.py
+++ b/substratest/factory.py
@@ -183,7 +183,7 @@ class DatasetSpec(_Spec):
     data_opener: str
     type: str
     description: str
-    metadata: typing.Dict = None
+    metadata: typing.Dict[str, str] = None
     permissions: Permissions = None
     objective_key: str = None
 
@@ -203,7 +203,7 @@ class ObjectiveSpec(_Spec):
     metrics: str
     test_data_sample_keys: typing.List[str]
     test_data_manager_key: str = None
-    metadata: typing.Dict = None
+    metadata: typing.Dict[str, str] = None
     permissions: Permissions = None
 
 
@@ -211,7 +211,7 @@ class _AlgoSpec(_Spec):
     name: str
     description: str
     file: str
-    metadata: typing.Dict = None
+    metadata: typing.Dict[str, str] = None
     permissions: Permissions = None
 
 
@@ -233,7 +233,7 @@ class TraintupleSpec(_Spec):
     train_data_sample_keys: typing.List[str]
     in_models_keys: typing.List[str] = None
     tag: str = None
-    metadata: typing.Dict = None
+    metadata: typing.Dict[str, str] = None
     compute_plan_id: str = None
     rank: int = None
 
@@ -243,7 +243,7 @@ class AggregatetupleSpec(_Spec):
     worker: str
     in_models_keys: typing.List[str]
     tag: str = None
-    metadata: typing.Dict = None
+    metadata: typing.Dict[str, str] = None
     compute_plan_id: str = None
     rank: int = None
 
@@ -255,7 +255,7 @@ class CompositeTraintupleSpec(_Spec):
     in_head_model_key: str = None
     in_trunk_model_key: str = None
     tag: str = None
-    metadata: typing.Dict = None
+    metadata: typing.Dict[str, str] = None
     compute_plan_id: str = None
     out_trunk_model_permissions: PrivatePermissions
     rank: int = None
@@ -267,7 +267,7 @@ class TesttupleSpec(_Spec):
     tag: str = None
     data_manager_key: str = None
     test_data_sample_keys: typing.List[str] = None
-    metadata: typing.Dict = None
+    metadata: typing.Dict[str, str] = None
 
 
 class ComputePlanTraintupleSpec(_Spec):
@@ -277,7 +277,7 @@ class ComputePlanTraintupleSpec(_Spec):
     traintuple_id: str
     in_models_ids: typing.List[str] = None
     tag: str = None
-    metadata: typing.Dict = None
+    metadata: typing.Dict[str, str] = None
 
     @property
     def id(self):
@@ -290,7 +290,7 @@ class ComputePlanAggregatetupleSpec(_Spec):
     worker: str
     in_models_ids: typing.List[str] = None
     tag: str = None
-    metadata: typing.Dict = None
+    metadata: typing.Dict[str, str] = None
 
     @property
     def id(self):
@@ -305,7 +305,7 @@ class ComputePlanCompositeTraintupleSpec(_Spec):
     in_head_model_id: str = None
     in_trunk_model_id: str = None
     tag: str = None
-    metadata: typing.Dict = None
+    metadata: typing.Dict[str, str] = None
     out_trunk_model_permissions: Permissions
 
     @property
@@ -317,7 +317,7 @@ class ComputePlanTesttupleSpec(_Spec):
     objective_key: str
     traintuple_id: str
     tag: str
-    metadata: typing.Dict = None
+    metadata: typing.Dict[str, str] = None
 
 
 def _get_key(obj, field='key'):
--- a/substratest/factory.py
+++ b/substratest/factory.py
@@ -388,9 +388,8 @@ class _BaseComputePlanSpec(_Spec, abc.ABC):
             train_data_sample_keys=_get_keys(data_samples),
             in_head_model_id=in_head_model.id if in_head_model else None,
             in_trunk_model_id=in_trunk_model.id if in_trunk_model else None,
-            out_trunk_model_permissions=out_trunk_model_permissions or DEFAULT_OUT_MODEL_PERMISSIONS,
-            tag=tag,
-            metadata=metadata
+            out_trunk_model_permissions=out_trunk_model_permissions or DEFAULT_OUT_TRUNK_MODEL_PERMISSIONS,
+            tag=tag
         )
         self.composite_traintuples.append(spec)
         return spec
--- a/substratest/factory.py
+++ b/substratest/factory.py
@@ -277,7 +277,6 @@ class ComputePlanTraintupleSpec(_Spec):
     traintuple_id: str
     in_models_ids: typing.List[str] = None
     tag: str = None
-    metadata: typing.Dict[str, str] = None
 
     @property
     def id(self):
@@ -290,7 +289,6 @@ class ComputePlanAggregatetupleSpec(_Spec):
     worker: str
     in_models_ids: typing.List[str] = None
     tag: str = None
-    metadata: typing.Dict[str, str] = None
 
     @property
     def id(self):
@@ -305,7 +303,6 @@ class ComputePlanCompositeTraintupleSpec(_Spec):
     in_head_model_id: str = None
     in_trunk_model_id: str = None
     tag: str = None
-    metadata: typing.Dict[str, str] = None
     out_trunk_model_permissions: Permissions
 
     @property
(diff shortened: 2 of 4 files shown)


substratest/assets.py
```diff
@@ -194,13 +194,15 @@ class DataSample(_Asset):
 class ObjectiveDataset(_InternalStruct):
     dataset_key: str
     data_sample_keys: typing.List[str]
+    metadata: typing.Dict
```
jmorel: The ObjectiveDataset isn't really an asset, it's a part of the Objective. It won't have metadata attached to it.

substratest/assets.py
```diff
@@ -194,13 +194,15 @@ class DataSample(_Asset):
...
 
 
 class Dataset(_Asset):
     key: str
     name: str
     owner: str
     objective_key: str
+    metadata: typing.Dict
```
jmorel: Since you specified in the backend that metadata is a dict of strings, you could also specify it here with the syntax typing.Dict[str, str] (keys are str, values are str)

jmorel: This would also apply to other assets.

substratest/assets.py
```diff
@@ -240,6 +244,7 @@ class TesttupleDataset(_InternalStruct):
     perf: float
     keys: typing.List[str]
     worker: str
+    metadata: typing.Dict
```
jmorel: As for the ObjectiveDataset, this isn't an Asset and won't have metadata.

substratest/factory.py
```diff
@@ -184,6 +184,7 @@ class DatasetSpec(_Spec):
     data_opener: str
     type: str
     description: str
+    metadata: typing.Dict
```
jmorel: As for assets.py, you could be more precise here and use `typing.Dict[str,str]`

substratest/factory.py
```diff
@@ -346,21 +357,24 @@ class _BaseComputePlanSpec(_Spec, abc.ABC):
     aggregatetuples: typing.List[ComputePlanAggregatetupleSpec]
     testtuples: typing.List[ComputePlanTesttupleSpec]
 
-    def add_traintuple(self, algo, dataset, data_samples, in_models=None, tag=''):
+    def add_traintuple(self, algo, dataset, data_samples, in_models=None, tag='', metadata=None):
         in_models = in_models or []
+        metadata = metadata or {}
```
jmorel: This line does the same as L370, you'll need to use one or the other!

jmorel: This also applies to add_aggregatuple and add_composite_traintuple.

jmorel: I suggest you use everywhere the pattern you're already using for add_testtuple.

substratest/factory.py
```diff
@@ -458,7 +476,8 @@ def create_data_sample(self, content=None, datasets=None, test_only=False):
             data_manager_keys=[d.key for d in datasets],
         )
 
-    def create_dataset(self, objective=None, permissions=None):
+    def create_dataset(self, objective=None, permissions=None, metadata=None):
+        metadata = metadata or {}
```
jmorel: This doubles with L501.

substratest/factory.py
```diff
@@ -479,12 +498,14 @@ def create_dataset(self, objective=None, permissions=None):
...
             type='Test',
+            metadata=metadata or {},
             description=str(description_path),
             objective_key=objective.key if objective else None,
             permissions=permissions or DEFAULT_PERMISSIONS,
         )
 
-    def create_objective(self, dataset=None, data_samples=None, permissions=None):
+    def create_objective(self, dataset=None, data_samples=None, permissions=None, metadata=None):
+        metadata = metadata or {}
```
jmorel: The pattern you're using here is different from the one you used for compute plan testtuples. You should chose one and apply it throughout the file.

substratest/assets.py
maeldebon: Yes, indeed!

substratest/factory.py
maeldebon: You're right, I forgot to remove it :/

substratest/assets.py
```diff
@@ -331,6 +337,7 @@ class Testtuple(_Asset, _FutureMixin):
     certified: bool
     rank: int
     tag: str
+    metadata: typing.Dict[str, str] = None
```
jmorel, line 340: The main description says testtuples are tested, but there's still `= None` here. Could you remove it?

substratest/factory.py
```diff
@@ -371,13 +383,14 @@ def add_aggregatetuple(self, aggregate_algo, worker, in_models=None, tag=''):
             worker=worker,
             in_models_ids=[t.id for t in in_models],
             tag=tag,
+            metadata=metadata or {}
```
jmorel, line 386: In half the cases, you've been doing `metadata=metadata` and in the other half `metadata=metadata or {}`. You should chose one of the two solutions and stick with it. When you do, could you also explain why you chose one over the other?

tests/test_network.py
```diff
@@ -17,10 +17,20 @@ def test_connection_to_nodes(clients):
...
 
+#
+# def test_add_dataset_with_metadata(factory, client):
+#     spec = factory.create_dataset(metadata={'foo': 'bar'})
+#     dataset = client.add_dataset(spec)
+#     assert dataset.metadata == {'foo': 'bar'}
+#
+#     dataset_copy = client.get_dataset(dataset.key)
+#     assert dataset == dataset_copy
+
```
jmorel, line 33: Could you cleanup these comments?

tests/test_network.py
```diff
@@ -124,6 +134,159 @@ def test_add_objective(factory, client):
...
+    too_long = 'foo' * 40
+    # raise an error if the key is too long (more than 50 chars)
+    spec = method_create(metadata={too_long: "bar"})
+    with pytest.raises(substra.exceptions.InvalidRequest):
+        method_add(spec)
+
+    # raise an error if the value is too long (more than 100 chars)
+    spec = method_create(metadata={"foo": too_long})
+    with pytest.raises(substra.exceptions.InvalidRequest):
+        method_add(spec)
```
jmorel, line 183: I feel like all the errors should be handled in the same way: always raise an InvalidRequest. Do you think it would be possible to update the backend and sdk code so that we have a consistent behavior from the user standpoint?

tests/test_network.py
```diff
@@ -124,6 +134,159 @@ def test_add_objective(factory, client):
...
+
+def test_metadata_aggregatetuple(factory, client, default_dataset):
+    composite_algo_spec = factory.create_composite_algo()
+    composite_algo = client.add_composite_algo(composite_algo_spec)
+    composite_traintuple_spec = factory.create_composite_traintuple(
+        algo=composite_algo,
+        dataset=default_dataset,
+        data_samples=default_dataset.train_data_sample_keys,
+    )
+    composite_traintuple = client.add_composite_traintuple(composite_traintuple_spec)
```
jmorel, line 272: You can create an aggregatetuple without inmodels (see the discussion at https://github.com/SubstraFoundation/substra-backend/pull/189). This would simplify this test a great deal.

tests/test_network.py
```diff
@@ -124,6 +134,159 @@ def test_add_objective(factory, client):
...
+    'aggregate_algo',
+    'composite_algo',
+])
+def test_metadata(factory, client, asset_name):
+    method_create = getattr(factory, f"create_{asset_name}")
+    method_add = getattr(client, f"add_{asset_name}")
+
+    # add an asset with metadata
+    spec = method_create(metadata={"foo": "bar"})
+    asset = method_add(spec)
```
jmorel, line 149: Since this test is trying out multiple metadata values, we could parametrize the metadata as well.

There could be another layer of parametrization to specify what the input metadata value, what the output should and which error should be raised.

Here is an example of combining parametrization:

```python
@pytest.mark.parametrize('y', (1, 2))
@pytest.mark.parametrize('x', ('a', 'b'))
def test_foo(x, y):
    assert True
```
This will be executed in the following order:
- a-1
- a-2
- b-1
- b-2

See also for how to handle exceptions https://github.com/SubstraFoundation/substra-backend/pull/189

jmorel:
When you'll be testing it for compute plans, don't forget that you'll also need to write tests for all compute plan assets.

tests/test_network.py
```diff
@@ -124,6 +134,159 @@ def test_add_objective(factory, client):
...
+    spec = method_create(metadata={"foo": too_long})
+    with pytest.raises(substra.exceptions.InvalidRequest):
+        method_add(spec)
+
+
+@pytest.mark.parametrize('asset_name, algo_type', [
+    ('traintuple', 'algo'),
+    ('composite_traintuple', 'composite_algo'),
+])
+def test_metadata_traintuple(factory, client, asset_name, algo_type, default_dataset):
```
jmorel, line 190: This test creates a lot of tuples. These tuples should be canceled if possible and the test should not end until the tuples are either done or canceled. Otherwise the execution of tuples in other tests might be delayed by these ones and timeout.

substratest/factory.py
```diff
@@ -179,6 +179,7 @@ class DatasetSpec(_Spec):
     data_opener: str
     type: str
     description: str
+    metadata: typing.Dict = None
```
jmorel, line 182: You could even go further and specify `typing.Dict[str, str] = None`, but this would also require to change the test on `metadata={'foo': {'bar': 'baz'}}` in the same way you did for `metadata='foo'`

substratest/factory.py
```diff
@@ -379,15 +392,17 @@ def add_composite_traintuple(self, composite_algo, dataset=None, data_samples=No
             in_trunk_model_id=in_trunk_model.id if in_trunk_model else None,
             out_trunk_model_permissions=out_trunk_model_permissions or DEFAULT_OUT_MODEL_PERMISSIONS,
             tag=tag,
+            metadata=metadata or {}
```
jmorel, line 395: Same as above.

substratest/factory.py
```diff
@@ -379,15 +392,17 @@ def add_composite_traintuple(self, composite_algo, dataset=None, data_samples=No
...
         self.composite_traintuples.append(spec)
         return spec
 
-    def add_testtuple(self, objective, traintuple_spec, tag=None):
+    def add_testtuple(self, objective, traintuple_spec, tag=None, metadata=None):
         spec = ComputePlanTesttupleSpec(
             objective_key=objective.key,
             traintuple_id=traintuple_spec.id,
             tag=tag or '',
+            metadata=metadata or {}
```
jmorel, line 405: Same as above.

tests/test_network.py
```diff
@@ -17,10 +17,20 @@ def test_connection_to_nodes(clients):
 def test_add_dataset(factory, client):
     spec = factory.create_dataset()
     dataset = client.add_dataset(spec)
+    # assert dataset.metadata == {}
```
jmorel, line 20: I feel this shouldn't be commented at all.

tests/test_network.py
```diff
@@ -124,6 +134,162 @@ def test_add_objective(factory, client):
...
+    spec = factory.create_aggregatetuple(
+        algo=aggregate_algo,
+        worker='MyOrg1MSP',
+        traintuples=[composite_traintuple],
+        metadata={"foo": "bar"}
+    )
+    asset = client.add_aggregatetuple(spec)
+
+    print(asset)
+    print(composite_traintuple)
```
jmorel, line 288: These should be removed.

tests/test_network.py
```diff
@@ -124,6 +134,162 @@ def test_add_objective(factory, client):
     assert dataset.objective_key == objective.key
 
 
+@pytest.mark.parametrize('asset_name', [
+    'objective',
+    'algo',
+    'aggregate_algo',
+    'composite_algo',
+])
+def test_metadata(factory, client, asset_name):
```
jmorel, line 143: Since this test is trying out multiple metadata values, we could parametrize the metadata as well.

There could be another layer of parametrization to specify what the input metadata value, what the output should and which error should be raised.

Here is an example of combining parametrization:
```python
@pytest.mark.parametrize('y', (1, 2))
@pytest.mark.parametrize('x', ('a', 'b'))
def test_foo(x, y):
    assert True
```
This will be executed in the following order:

* a-1
* a-2
* b-1
* b-2

See also for how to handle exceptions SubstraFoundation/substra-backend#189

tests/test_network.py
```diff
@@ -124,6 +128,102 @@ def test_add_objective(factory, client):
...
+    #'algo',
+    #'aggregate_algo',
+    #'composite_algo',
+])
+@pytest.mark.parametrize('metadata,metadata_output,metadata_error', [
+    ({'foo': 'bar'}, {'foo': 'bar'}, {'foo': {"bar": "foo"}}),
+    (None, {}, {too_long: "bar"}),
+    ({}, {}, {"foo": too_long}),
+])
+def test_metadata(factory, client, asset_name, metadata, metadata_output, metadata_error):
```
jmorel, line 143: This test could be splitted in 2 tests: test_valid_metadata and test_invalid_metadata. The former one would use the asset_name / metadata / metadata_output as args, and the latter one would use just asset_name and metadata. 

In order to handle all types of bad metadata in the same way, you would have to set the bad metadata on the spec_dict directly.

tests/test_network.py
```diff
@@ -124,6 +128,102 @@ def test_add_objective(factory, client):
...
+    ('traintuple', 'algo'),
+    ('composite_traintuple', 'composite_algo'),
+])
+@pytest.mark.parametrize('metadata,metadata_output,metadata_error', [
+    ({'foo': 'bar'}, {'foo': 'bar'}, {'foo': {"bar": "foo"}}),
+    (None, {}, {too_long: "bar"}),
+    ({}, {}, {"foo": too_long}),
+])
+def test_metadata_traintuple(factory, client, asset_name, algo_type, default_dataset,
+                             metadata, metadata_output, metadata_error):
```
jmorel, line 176: As for test_metadata, this test could be split in 2.

tests/test_network.py
```diff
@@ -124,6 +128,102 @@ def test_add_objective(factory, client):
...
+        dataset=default_dataset,
+        data_samples=default_dataset.train_data_sample_keys,
+    )
+    composite_traintuple = client.add_composite_traintuple(composite_traintuple_spec)
+
+    aggregate_algo_spec = factory.create_aggregate_algo()
+    aggregate_algo = client.add_aggregate_algo(aggregate_algo_spec)
+    spec = factory.create_aggregatetuple(
+        algo=aggregate_algo,
+        worker='MyOrg1MSP',
```
jmorel, line 218: The only hardcoded values in tests should be the values you're actually testing. In this case, The worker is `MyOrg1MSP` because it's the value of client.node_id, so you should use `client.node_id` instead.

tests/test_network.py
```diff
@@ -124,6 +128,102 @@ def test_add_objective(factory, client):
...
+        traintuples=[composite_traintuple],
+        metadata={"foo": "bar"}
```
jmorel, line 220: The test should be parametrized to test good values and bad values.

tests/test_network.py
```diff
@@ -124,6 +128,102 @@ def test_add_objective(factory, client):
     assert dataset.objective_key == objective.key
 
 
+@pytest.mark.parametrize('asset_name', [
+    'dataset',
+    #'objective',
+    #'algo',
+    #'aggregate_algo',
+    #'composite_algo',
+])
```
jmorel, line 137: It looks like you commented the wrong lines here :)

substratest/factory.py
```diff
@@ -179,6 +179,7 @@ class DatasetSpec(_Spec):
     data_opener: str
     type: str
     description: str
+    metadata: typing.Dict[str, str] = None
```
samlesu, line 182: We could do the same and add an another `_Metadata` type for this file.

tests/test_network.py
```diff
@@ -124,6 +128,154 @@ def test_add_objective(factory, client):
...
+    'algo',
+    'aggregate_algo',
+    'composite_algo',
+])
+@pytest.mark.parametrize('metadata,metadata_output', [
+    ({'foo': 'bar'}, {'foo': 'bar'}),
+    (None, {}),
+    ({}, {}),
+])
+def test_metadata(factory, client, asset_name, metadata, metadata_output):
```
samlesu, line 143: Could be named test_add_metadata to be consistent with other test names.

tests/test_network.py
```diff
@@ -124,6 +128,154 @@ def test_add_objective(factory, client):
...
+    'aggregate_algo',
+    'composite_algo',
+])
+@pytest.mark.parametrize('metadata', [
+    'foo',
+    {'foo': {"bar": "foo"}},
+    {too_long: "bar"},
+    {"foo": too_long},
+])
+def test_invalid_metadata(factory, client, asset_name, metadata):
```
samlesu, line 166: To be rename accordingly with the other test.

substratest/assets.py
```diff
@@ -201,6 +201,7 @@ class Dataset(_Asset):
     name: str
     owner: str
     objective_key: str
+    metadata: typing.Dict[str, str]
```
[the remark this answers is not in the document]
maeldebon: It doesn't work :/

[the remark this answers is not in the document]
samlesu: Thanks for investigating. Let's keep the current version, this could be improved later.

substratest/factory.py
samlesu: Could be investigated later.

substratest/assets.py
```diff
@@ -293,8 +297,9 @@ class Aggregatetuple(_Asset, _FutureMixin):
     compute_plan_id: str
     rank: int
     tag: str
+    metadata: typing.Dict[str, str]
     log: str
-    in_models: typing.List[InModel]
+    in_models: typing.List[InModel] = None
```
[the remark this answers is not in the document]
samlesu: @maeldebon Could you answer this comment?

tests/test_network.py
```diff
@@ -124,6 +125,106 @@ def test_add_objective(factory, client):
...
+@pytest.mark.parametrize('asset_name,algo_type', [
+    ('traintuple', 'algo'),
+    ('composite_traintuple', 'composite_algo'),
+])
+@pytest.mark.parametrize('metadata,metadata_output', [
+    ({'foo': 'bar'}, {'foo': 'bar'}),
+    (None, {}),
+    ({}, {}),
+])
+def test_traintuple_with_metadata(factory, client, asset_name, algo_type, default_dataset, metadata, metadata_output):
```
samlesu, line 180: As said in a previous comment I think we should remove this test as well as the next one (`test_traintuple_with_invalid_metadata`).

In my opinion, the end to end tests should not test all the use cases and all the corner cases. This should be done in the unit test. So for maintainability and performance reasons, I think they are not needed.

substratest/factory.py
```diff
@@ -329,7 +336,7 @@ class _BaseComputePlanSpec(_Spec, abc.ABC):
     aggregatetuples: typing.List[ComputePlanAggregatetupleSpec]
     testtuples: typing.List[ComputePlanTesttupleSpec]
 
-    def add_traintuple(self, algo, dataset, data_samples, in_models=None, tag=''):
+    def add_traintuple(self, algo, dataset, data_samples, in_models=None, tag='', metadata=None):
```
jmorel, line 339: These `add_<asset>` are creating compute plan specs. It's been decided to handle compute plan specs in a separate PR, so they should be removed.

substratest/assets.py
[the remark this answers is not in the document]
maeldebon: Yep, this asset is supposed to support being created without in_model keys, but it was never tested before. That's why I added this, to allow us to create it without these keys. 
But yes, this isn't directly related to the PR :/

substratest/factory.py
maeldebon: You're right, I'll do this!

jmorel:
You need to rebase your branch apparently.

substratest/factory.py
```diff
@@ -381,8 +388,9 @@ def add_composite_traintuple(self, composite_algo, dataset=None, data_samples=No
             train_data_sample_keys=_get_keys(data_samples),
             in_head_model_id=in_head_model.id if in_head_model else None,
             in_trunk_model_id=in_trunk_model.id if in_trunk_model else None,
-            out_trunk_model_permissions=out_trunk_model_permissions or DEFAULT_OUT_TRUNK_MODEL_PERMISSIONS,
+            out_trunk_model_permissions=out_trunk_model_permissions or DEFAULT_OUT_MODEL_PERMISSIONS,
```
samlesu, line 391: What's the purpose of this change?

PR #91, opened 2020-03-20
Allow dataset / data samples in testtuple spec

jmorel:
Enable the creation of testtuples not just from an objective but also through explicit dataset and data samples.

### DIFF
--- a/substratest/factory.py
+++ b/substratest/factory.py
@@ -269,6 +269,8 @@ class TesttupleSpec(_Spec):
     objective_key: str
     traintuple_key: str
     tag: str
+    data_manager_key: str = None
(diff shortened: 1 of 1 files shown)


AurelienGasser:
@jmorel did you run the tests to try it?

jmorel:
I just ran the tests right now @AurelienGasser and I can confirm that even if I pass [] instead of None for the data samples, the test data samples from the objectives are still used.

This was confirmed by the testtuple created in the `test_tuples_execution_on_same_node` test. The testtuple content is:
```yaml
```

AurelienGasser:
@jmorel thanks for checking!

PR #92, opened 2020-03-24
Remove public key from trunk model permissions

jmorel:
Companion PR https://github.com/SubstraFoundation/substra-backend/pull/188

Remove public key from out composite traintuple specs.

### DIFF
--- a/tests/test_permissions.py
+++ b/tests/test_permissions.py
@@ -314,30 +314,3 @@ def test_permissions_denied_head_model_process(factory, client_1, client_2):
     )
     with pytest.raises(substra.exceptions.InvalidRequest):
         client_2.add_composite_traintuple(spec)
-
-
-def test_permission_public_trunk(factory, client):
-    spec = factory.create_dataset()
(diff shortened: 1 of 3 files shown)


jmorel:
@AurelienGasser @samlesu could you re-review this PR ? I added a test making sure you cannot force public permissions for a composite traintuple trunk.

tests/test_permissions.py
```diff
@@ -314,3 +314,30 @@ def test_permissions_denied_head_model_process(factory, client_1, client_2):
     )
     with pytest.raises(substra.exceptions.InvalidRequest):
         client_2.add_composite_traintuple(spec)
+
+
+def test_permission_public_trunk(factory, client):
```
samlesu, line 319: It is not really clear what's the purpose of this test. Adding a docstring/comments would be IMO a good idea.

jmorel:
@samlesu done, I moved it to the system tests and added an explanation.

samlesu:
@jmorel thanks!

PR #95, opened 2020-04-02
Test data samples relative order in traintuples

jmorel:
Adds a test checking that the relative order of data samples in traintuples is not preserved (yet). 

Since preserving this relative order is something we'll soon want to do, I opted for writing tests that check that the order is preserved and making them as failing.

### DIFF
--- a/tests/test_data_samples_order.py
+++ b/tests/test_data_samples_order.py
@@ -3,7 +3,7 @@ import pytest
 import copy
 
 
-ALGO_SCRIPT = """
+TEMPLATE_ALGO_SCRIPT = """
 import json
 import substratools as tools
 class TestAlgo(tools.Algo):
@@ -23,7 +23,7 @@ if __name__ == '__main__':
     tools.algo.execute(TestAlgo())
 """
 
-COMPOSITE_ALGO_SCRIPT = """
+TEMPLATE_COMPOSITE_ALGO_SCRIPT = """
 import json
 import substratools as tools
 class TestCompositeAlgo(tools.CompositeAlgo):
@@ -65,7 +65,7 @@ def _shuffle(items):
 def test_traintuple_data_samples_relative_order(factory, client, default_dataset):
     data_sample_keys = _shuffle(default_dataset.train_data_sample_keys)
 
-    algo_script = ALGO_SCRIPT.format(data_sample_keys=data_sample_keys, models=None)
+    algo_script = TEMPLATE_ALGO_SCRIPT.format(data_sample_keys=data_sample_keys, models=None)
     spec = factory.create_algo(py_script=algo_script)
     algo = client.add_algo(spec)
 
@@ -75,6 +75,9 @@ def test_traintuple_data_samples_relative_order(factory, client, default_dataset
         data_samples=data_sample_keys,
     )
     traintuple = client.add_traintuple(spec)
+    # Ensure the order of the data sample keys is correct at 2 levels: :
+    #  1. In the returned traintuple
+    #  2. In the train method of the algo. If the order is incorrect, .future().wait() will fail.
     assert traintuple.dataset.keys == data_sample_keys
     traintuple.future().wait()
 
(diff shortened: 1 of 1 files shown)


tests/test_data_samples_order.py
```diff
@@ -0,0 +1,97 @@
...
+    spec = factory.create_algo(py_script=algo_script)
+    algo = client.add_algo(spec)
+
+    spec = factory.create_traintuple(
+        algo=algo,
+        dataset=default_dataset,
+        data_samples=data_sample_keys,
+    )
+    traintuple = client.add_traintuple(spec)
+    assert traintuple.dataset.keys == data_sample_keys
```
AurelienGasser, line 78:
```suggestion
    # Ensure the order of the dataset keys is correct at 2 levels: :
    #  1. In the returned traintuple
    #  2. In the train method of the algo. If the order is incorrect, .future().wait() will fail.
    assert traintuple.dataset.keys == data_sample_keys
```

tests/test_data_samples_order.py
```diff
@@ -0,0 +1,97 @@
...
+    spec = factory.create_composite_algo(py_script=composite_algo_script)
+    composite_algo = client.add_composite_algo(spec)
+
+    spec = factory.create_composite_traintuple(
+        algo=composite_algo,
+        dataset=default_dataset,
+        data_samples=data_sample_keys,
+    )
+    composite_traintuple = client.add_composite_traintuple(spec)
+    assert composite_traintuple.dataset.keys == data_sample_keys
```
AurelienGasser, line 96:
```suggestion
    # Ensure the order of the dataset keys is correct at 2 levels: :
    #  1. In the returned traintuple
    #  2. In the train method of the algo. If the order is incorrect, .future().wait() will fail.
    assert composite_traintuple.dataset.keys == data_sample_keys
```

tests/test_data_samples_order.py
```diff
@@ -0,0 +1,97 @@
...
+import copy
+
+
+ALGO_SCRIPT = """
+import json
+import substratools as tools
+class TestAlgo(tools.Algo):
+    def train(self, X, y, models, rank):
+        data_sample_keys = [folder.split('/')[-1] for folder in X]
+        assert data_sample_keys == {data_sample_keys}, data_sample_keys
```
samlesu, line 12: What's the purpose of this check?

As far as I know, this is comparing a list and a set, this will never be True.

tests/test_data_samples_order.py
```diff
@@ -0,0 +1,97 @@
...
+        dataset=default_dataset,
+        data_samples=data_sample_keys,
+    )
+    traintuple = client.add_traintuple(spec)
+    assert traintuple.dataset.keys == data_sample_keys
+    traintuple.future().wait()
+
+
+@pytest.mark.xfail(reason='data samples relative order is not preserved (yet)')
+def test_composite_traintuple_data_samples_relative_order(factory, client, default_dataset):
```
samlesu, line 83: IMO a single test on the traintuple is sufficient, we should expect the same behaviour for all tuples. But that's really minor.

tests/test_data_samples_order.py
```diff
@@ -0,0 +1,97 @@
...
+import pytest
+import copy
+
+
+ALGO_SCRIPT = """
+import json
+import substratools as tools
+class TestAlgo(tools.Algo):
+    def train(self, X, y, models, rank):
+        data_sample_keys = [folder.split('/')[-1] for folder in X]
```
samlesu, line 11: I'm not 100% sure you should base the test on the folder name that is currently equal to the data samples hash. This could change later.

IMO the best from a testing perspective would be to load the data values and check the order is as expected for instance. But not sure this is easy to do with the current substra-tests implementation.

tests/test_data_samples_order.py
jmorel: It's not actually a set, it's an interpolated variable. It is replaced by the actual list of data sample keys in the tests themselves.

tests/test_data_samples_order.py
jmorel: I went for the most straightforward solution for now, but I agree that in the future we should fond a more durable solution. But as long as we don't have a mechanism that enforces the relative order I'm not sure it's worth the hassle.

samlesu: Sure!

tests/test_data_samples_order.py
samlesu: Oh got it!
To avoid confusion what about:
- using template string (https://docs.python.org/3/library/string.html#template-strings),
- or, renaming the variable to `TEMPLATE_ALGO_SCRIPT`

jmorel:
@samlesu @AurelienGasser done, thanks!

PR #99, opened 2020-04-20
CI: Use 'eu' container registry

AurelienGasser:
- Store docker images in the same gcloud zone as the k8s cluster (`eu`)
  - This minimizes the $ cost of network transfers between the container registry and the kubernetes cluster (see [gcloud pricing](https://cloud.google.com/storage/pricing#network-egress))
- Add a `ci-` prefix to generated docker image tags. 
  - This makes it explicit that those images were created for the purpose of being run as part of CI script, and they can safely be deleted after the CI run has finished.
﻿


AurelienGasser:
Tested `run-ci.py`: "it works" :+1:

PR #100, opened 2020-04-23
Move system tests to a separate module

### LINKED ISSUE #87: Separate system vs user tests
Move test checking system features into a dedicated module (celery retry, delete intermediary models, ...).

jmorel:
Fixes #87 

This creates a new `test_system.py` module that contains tests for internal features of the platform.

Moved tests:
* test_execution_retry_on_fail

Tests not moved yet (to be debated):
* test_compute_plan_remove_intermediary_model
* test_add_data_sample_located_in_shared_path
* test_traintuple_data_samples_relative_order 
* test_composite_traintuple_data_samples_relative_order

These 4 tests can be seen as both system and non system at the same time, we just need to decide in which category they should go.
@samlesu, since this was you issue what would you like to do for each of these tests?

### DIFF
--- a/tests/test_execution.py
+++ b/tests/test_execution.py
@@ -6,7 +6,6 @@ import substratest as sbt
 from substratest.factory import Permissions
 
 from substratest import assets
-from . import settings
 
 
 @pytest.mark.slow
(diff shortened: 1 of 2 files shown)


samlesu:
Thanks @jmorel !

It's indeed debatable. It looks good to me.

PR #102, opened 2020-04-28
CI: display args at the end of the run too

AurelienGasser:


ci/run-ci.py
```diff
@@ -137,8 +137,11 @@ def arg_parse():
     if args['no_cache']:
         KANIKO_CACHE_TTL = '-1h'
 
+    print(f'💃💃💃\n')
```
jmorel, line 140: I think there's no need for the `\n ` here, there will be a line return automatically.

AurelienGasser: Thanks! It turns out the newline looks good in this context, so I'll just leave it.

PR #105, opened 2020-05-05
Test default permissions

jmorel:
Companion PR to https://github.com/SubstraFoundation/substra-backend/pull/247

Make sure that assets are created with private default permissions.

### DIFF
--- a/tests/test_system.py
+++ b/tests/test_system.py
@@ -101,3 +101,70 @@ def test_permission_public_trunk(factory, client):
     composite_traintuple = sbt.assets.CompositeTraintuple.load(res).attach(client)
     assert composite_traintuple.out_trunk_model.permissions.process.public is False
     composite_traintuple.future().wait()
+
+
+@pytest.mark.parametrize('include_permissions_key', (True, False))
+@pytest.mark.parametrize('asset_name,asset_class', (
+    ('dataset', sbt.assets.Dataset),
+    ('objective', sbt.assets.Objective),
+    ('algo', sbt.assets.Algo),
+    ('composite_algo', sbt.assets.CompositeAlgo),
+    ('aggregate_algo', sbt.assets.AggregateAlgo),
+))
+def test_permission_defaults(factory, client, asset_name, asset_class, include_permissions_key):
+    create_method = getattr(factory, f'create_{asset_name}')
+    add_method = getattr(client._client, f'add_{asset_name}')
+
+    spec = create_method()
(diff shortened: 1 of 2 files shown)


jmorel:
@AurelienGasser @inalgnu @samlesu This is ready for review, could you have a look at it ? The tests are passing.

tests/test_system.py
```diff
@@ -101,3 +101,70 @@ def test_permission_public_trunk(factory, client):
...
+    ('aggregate_algo', sbt.assets.AggregateAlgo),
+))
+def test_permission_defaults(factory, client, asset_name, asset_class, include_permissions_key):
+    create_method = getattr(factory, f'create_{asset_name}')
+    add_method = getattr(client._client, f'add_{asset_name}')
+
+    spec = create_method()
+    spec_dict = spec.dict()
+    if include_permissions_key:
+        spec_dict['permissions'] = None
```
AurelienGasser, line 121: What's the rationale for setting it to `None`?

AurelienGasser: Do we want to keep on supporting "empty" permissions in the future?
Do we  want to keep on supporting "missing" permissions in the future?

jmorel: My reasoning is that we want to have a proper default for permissions, and that we should therefore accept all forms of non-specifified permissions (both empty and missing).

jmorel: As for the future, I'll let the powers that be decide!

AurelienGasser: I would be in favor of only having a test for the "missing" case in `substra-tests`. The rationale being that it's the usage we recommend and document.

If we also want to support the "empty" case, then I think adding a unit test in `substra-backend` would be more appropriate.

What do you think?

AurelienGasser: Also please note that I'm being nitpicky here, if there's any urgency to this PR I'm happy to mark it as approved :)

samlesu: Substra-tests has not been designed for testing system features / behaviours (substra-tests does not provide the proper tools to do that). In my opinion, we should aim at reducing the number of system tests not increasing it. So if we can write unit tests for this PR, I think it would be more appropriate.

samlesu: Regarding the support of different permissions format, I think we should support only the specified forms of permissions (I prefer to be more restrictive).

jmorel:
Closing this as examples now have explicit permissions set.

PR #107, opened 2020-05-07
Use latest kaniko version that fix gcloud build authentication

Kelvin-M:


Kelvin-M:
Tested with `run-ci.py` no issue ! :)

AurelienGasser:
Thanks @Kelvin-M !

PR #116, opened 2020-07-09
Add n_fake_samples and more realistic samples and output to the factory

Esadruhn:
### Summary

Update substra tests to use the latest version of substra tools (fake methods take n_samples as argument) and generate more realistic output in the factory. This may allow us to test that everything went well in the workflow by asserting that the final performance equals a certain value.

### Main features

- update the factory opener script, methods fake_X and fake_y take n_samples as argument and their output depends on its value
- update the factory opener script, get_X, get_y and the data sample generator for more realistic outputs

### Extra

When the https://github.com/SubstraFoundation/substra/pull/195 is merged, it will be possible to use "fake_data" from the sdk so we will be able to test that.

### DIFF
--- a/substratest/factory.py
+++ b/substratest/factory.py
@@ -1,5 +1,4 @@
 import abc
-import json
 import os
 import pathlib
 import random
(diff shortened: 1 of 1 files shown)


Esadruhn:
Reverted the changes on the default assets, there are already 2 PRs on realistic output

PR #117, opened 2020-07-13
Make tests run on the local and remote backend

Esadruhn:
Some tests run on the local and remote backend, some only on the remote and some only on the local

## Main changes

- Possibility of choosing the backend the tests run on with a command line option (default `remote`): 
  - `pytest --local` to run on the local backend, 
  - no `local` flag to run on the remote backend
- Mark a test to run only 
  - on the local backend: `pytest.mark.local_only`
  - on the remote backend: `pytest.mark.remote_only`
- Makefile: 
  - make test --> run the tests on the remote backend
  - make test-local --> run the tests on the local backend
- Fixtures: the remote backend has two nodes and the local backend one, so it is not possible to use the 'client_2', 'dataset_2', ... fixtures with local tests. There may be unexpected behaviour (ie fail) on the tests that use the 'clients', 'default_datasets'... fixtures and expects two clients with different names.

## Merge request requirements

- [x] merge https://github.com/SubstraFoundation/substra/pull/197 before, so that the tests on the aggregate / composite traintuple pass
- [ ] merge https://github.com/SubstraFoundation/substra/pull/202 to make the tests on the compute plan run on the local backend

An alternative is to mark the failing tests as 'remote_only'

## N.B.

Following this, remove the tests on the local backend from substra.


Esadruhn:
Question: what does network.enable_intermediate_model_removal do ?

README.md
```diff
@@ -20,7 +20,7 @@ It can be found [here](https://github.com/SubstraFoundation/substra)
 
 You will need to install it thanks to the `pip` binary.
 
-# Run the tests
+# Run the tests on the default backend
```
jmorel, line 23: I'd say "on a remote backend" since the section provides explanation for both default and alternate remote backends.

README.md
```diff
@@ -52,6 +52,21 @@ Note that `test_compute_plan` from `test_execution_compute_plan.py` is not marke
...
+
+By default the tests (default and minimal) run on the default backend: a running Substra platform.
+They can also be run on the local backend (no Substra deployment).
+
+The network configuration is described in a yaml file: `values_local_backend.yaml` and cannot be changed.
+
+To run the tests using on the local backend:
+
+```
+test-local
```
jmorel, line 65: `make test-local` :)

README.md
```diff
@@ -52,6 +52,21 @@ Note that `test_compute_plan` from `test_execution_compute_plan.py` is not marke
...
+```
+
+Some tests are skipped in this mode as they need the default backend to run.
```
jmorel, line 68: `as they need a remote backend to run.`

substratest/factory.py
```diff
@@ -429,7 +429,7 @@ def __init__(self, name):
         self._dataset_counter = Counter()
         self._objective_counter = Counter()
         self._algo_counter = Counter()
-        self._workdir = pathlib.Path(tempfile.mkdtemp())
+        self._workdir = pathlib.Path(tempfile.mkdtemp(prefix=os.getcwd() + '/'))
```
jmorel, line 432: Is it absolutely necessary to work in the CWD?

tests/conftest.py
```diff
@@ -28,6 +28,45 @@ def pytest_configure(config):
...
+    )
+
+
+def pytest_collection_modifyitems(config, items):
+    """Skip the remote tests if local backend and local tests if remote backend.
+    If both or none are specified, run all tests on both.
+    """
+    local = config.getoption("--local")
+    if local:
+        skip_marker = pytest.mark.skip(reason="need --remote option to run")
```
jmorel, line 54: There is no `--remote` option that I can see.

values_local_backend.yaml
```diff
@@ -0,0 +1,7 @@
+options:
```
jmorel, line 1: To keep with the current naming (`values` / `local-values`) you could name this file `local-backend-values.yaml`.

tests/test_execution.py
```diff
@@ -135,6 +137,7 @@ def test_traintuple_execution_failure(factory, client, default_dataset_1):
 
 
 @pytest.mark.slow
+@pytest.mark.remote_only
```
jmorel, line 140: I think this should work on local backend as well.

tests/test_execution.py
```diff
@@ -153,6 +156,7 @@ def test_composite_traintuple_execution_failure(factory, client, default_dataset
...
```
jmorel, line 159: I think this should work on local backend as well.

tests/test_execution.py
```diff
@@ -118,6 +119,7 @@ def test_tuples_execution_on_different_nodes(factory, client_1, client_2, defaul
...
```
jmorel, line 122: I think this should work on local backend as well.

tests/test_execution.py
Esadruhn: The tests on 'failure' do not work with the local backend since in the local backend an error is raised whereas in the remote the status is set to "failed".
I can adapt the test to `with pytest.raises` if it is on the local backend though

jmorel: Ah yes, I get it. Well, using pytest.raises here would be great!

jmorel:
This is looking pretty good!

substratest/factory.py
Esadruhn: The error if I add nothing is:
```
E       docker.errors.APIError: 500 Server Error: Internal Server Error ("b'Mounts denied: \r\nThe path /var/folders/k2/kkx936g94tn2y6xtrdf8cj_m0000gn/T/tmpbl4cf2dv/dataset-0/opener.py\r\nis not shared from OS X and is not known to Docker.\r\nYou can configure shared paths from Docker -> Preferences... -> File Sharing.\r\nSee https://docs.docker.com/docker-for-mac/osxfs/#namespaces for more info.\r\n.'")
```
For the Finder, /var/folders does not exist and without both /var/folders and /private/var/folders added to the shared files we get this error.

--> We could use `/tmp/` instead

jmorel: 👍

Esadruhn:
Thanks, I've been running the tests on the compute plan and I will add them to this PR if that's OK with you

ci/run-ci.py
```diff
@@ -425,6 +425,7 @@ def run_tests():
 
     try:
         call(f'kubectl --context {KUBE_CONTEXT} exec {substra_tests_pod} -n substra-tests -- make test')
+        call(f'kubectl --context {KUBE_CONTEXT} exec {substra_tests_pod} -n substra-tests -- make test-local')
```
AurelienGasser, line 428: For the CI we want to run the full end to end tests using the deployed backend

Esadruhn: With this the e2e tests run both on the deployed and mock backend, no ?

AurelienGasser: I think `make test-local` will only run the local tests (see my comment in `tests/conftest.py`)

I would prefer the CI to only run the "remote backend" tests: if a "local backend" test fails, i don't think we want this to trigger a CI failure, which in turns triggers a message in chat.

README.md
```diff
@@ -20,7 +20,7 @@ It can be found [here](https://github.com/SubstraFoundation/substra)
 
 You will need to install it thanks to the `pip` binary.
 
-# Run the tests
+# Run the tests on the remote backend
```
AurelienGasser, line 23: The "remote backend" name is misleading because it implies that the backend runs on a "remote" machine, which is often not the case, especially on local environments.

Can we consider a different naming scheme?

e.g.

- deployed backend / mock backend
- backend / mock backend
- something else?

tests/conftest.py
```diff
@@ -28,6 +28,45 @@ def pytest_configure(config):
...
+    parser.addoption(
+        "--local",
+        action="store_true",
+        help="Run the tests on the local backend only. Otherwise run the tests only on the remote backend."
+    )
+
+
+def pytest_collection_modifyitems(config, items):
+    """Skip the remote tests if local backend and local tests if remote backend.
+    If both or none are specified, run all tests on both.
```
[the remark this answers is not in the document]
Esadruhn: Indeed, the comment is outdated

ci/run-ci.py
Esadruhn: I would actually trigger a CI failure for tests on the local backend, as the local backend must be similar to the remote backend to be useful. Without a CI I feel that the local backend will become out-of-date quickly.

README.md
Esadruhn: I agree, "mock" backend is better than "local". Since it is also "local" in Substra, let's change all of it at once when the features are merged.

tests/settings.py
```diff
@@ -63,17 +65,35 @@ def load():
...
+
+
 # TODO that's a bad idea to expose the static configuration, it has been done to allow
 #      tests parametrization but this won't work for specific tests written with more
 #      nodes
 
 # load configuration at module load time to allow tests parametrization depending on
 # network static configuration
 load()
+load_local_backend()
```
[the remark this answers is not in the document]
Esadruhn: It does not seem to be, I am not sure why "local()" is needed too, I'll remove it.

ci/run-ci.py
Esadruhn: @jmorel What do you think ?

Esadruhn:
You are right, there should be a way to run the tests on the remote and local backend at the same time and devs should be aware that testing on both is required.
I changed the makefile commands: `make test` run the tests on both backends, use `make test-remote` and `make test-local` to choose the backend.

ci/run-ci.py
jmorel: I agree with @Esadruhn, I think of these e2e tests as a guarantee for the user experience and not just as validation that changes in the backend / chaincode did not break anything. And in order for the local backend to be useful, we must have it as close to the remote backend as possible.

Makefile
```diff
@@ -12,10 +12,17 @@ pyclean:
 
 test: pyclean
 	pytest tests -rs -v --durations=0
+	pytest tests -rs -v --durations=0 --local
```
jmorel, line 15: Since test is just test-remote + test-local, you could rewrite this as `test: test-remote test-local`. This way any change it the way we launch either set of tests is automatically included in the global test target.

README.md
```diff
@@ -22,6 +22,15 @@ You will need to install it thanks to the `pip` binary.
 
 # Run the tests
 
+The tests can run both on the deployed backend and the mock backend. To run the complete
+test suite on both backend:
```
jmorel, line 26: You're talking about deployed/mock here but the everywhere you use local/remote. I'd be for sticking with local/remote and then update the wording across all repos at once.

Makefile
Esadruhn: Nice, thanks !

PR #119, opened 2020-07-17
Tests on testtuple perf: replace equality test between float by isclose

Esadruhn:
## What this changes

In the tests, one common assertion is on the final performance of the testtuple:
`assert testtuple.dataset.perf == 0.2`

These tests fail at least with the local backend because of float precision:
`assert 0.20000000000000018 == 0.2`

The proposition is to replace all `assert testtuple.dataset.perf == x` with `assert math.isclose(testtuple.dataset.perf, x)`

### DIFF
--- a/tests/test_execution.py
+++ b/tests/test_execution.py
@@ -1,3 +1,4 @@
+import math
 import pytest
 
 import substra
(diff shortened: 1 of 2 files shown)


jmorel:
Thanks @Esadruhn I run into the issue as well when I'm using the local backend, but I don't remember having the issue with a remote backend, do you know why that is?

PR #122, opened 2020-07-21
Separate CI for the tests on the remote and local backend

Esadruhn:
Separate the CI on the remote and local backend:
- the tests on the remote are run with a deployed substra platform
- the tests on the local are run with only substra-tests and substra installed

Skip a test:
- skip the "shared paths" test on the local backend

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,9 @@ branches:
 git:
   depth: 1
 
+services:
+  - docker
+
 before_install:
   # Decrypt secrets
   # Secrets generated using the CLI: `travis encrypt-file --add [...]`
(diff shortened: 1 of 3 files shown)


Esadruhn:
- What is the use of defining the substra repo and branch in env variables ? Do we want to be able to change it from the code or when triggering a build from Travis ?

jmorel:
You can use run-ci manually and decide which branch of the backend, hlf-k8s and substra-test to use, which is very handy when validating a work in progress. Did you already ran the CI to check that your new travis.yml works as expected? The separate jobs are a very nice addition btw!

PR #123, opened 2020-07-22
Update with the hybrid debugging

Esadruhn:
## Main changes

- Update substra-tests to add tests on the debug mode (following the changes in Substra: https://github.com/SubstraFoundation/substra/pull/206)
- Make it easier to debug failures in performance assert: 
  - add print statements in the assets (these will be part of the execution logs in the local backend)
  - print the execution logs in the tests (not shown in case of success)
- Fixed the CI: upgrade Docker to version 19
  - I fixed the versions of docker-ce, docker-ce-cli, containerd, **do we want to always use the latest stable version instead ?**

## Requirements

Need this PR to be merged: https://github.com/SubstraFoundation/substra-tools/pull/42 and to update the substra-tools image

### DIFF
--- a/substratest/client.py
+++ b/substratest/client.py
@@ -21,8 +21,8 @@ class Client:
         self.debug = debug
         self._client = substra.Client(debug=debug, url=address, version="0.0", insecure=False, token=token)
         if not token:
-            self._client.login(user, password)
-        self.token = self._client._token
+            token = self._client.login(user, password)
+        self.token = token
 
     def add_data_sample(self, spec, *args, **kwargs):
         res = self._client.add_data_sample(spec.dict(), *args, **kwargs)
--- a/substratest/client.py
+++ b/substratest/client.py
@@ -14,13 +14,15 @@ class Client:
     Parses responses from server to return Asset instances.
     """
 
-    def __init__(self, debug, node_id=None, address=None, user=None, password=None):
+    def __init__(self, debug, node_id=None, address=None, token=None, user=None, password=None):
         super().__init__()
 
         self.node_id = node_id
         self.debug = debug
-        self._client = substra.Client(debug=debug, url=address, version="0.0", insecure=False)
-        self._client.login(user, password)
+        self._client = substra.Client(debug=debug, url=address, version="0.0", insecure=False, token=token)
+        if not token:
+            self._client.login(user, password)
+        self.token = self._client._token
 
     def add_data_sample(self, spec, *args, **kwargs):
         res = self._client.add_data_sample(spec.dict(), *args, **kwargs)
--- a/substratest/client.py
+++ b/substratest/client.py
@@ -14,12 +14,12 @@ class Client:
     Parses responses from server to return Asset instances.
     """
 
-    def __init__(self, backend, node_id=None, address=None, user=None, password=None):
+    def __init__(self, debug, node_id=None, address=None, user=None, password=None):
         super().__init__()
 
         self.node_id = node_id
-        self.backend = backend
-        self._client = substra.Client(backend=backend, url=address, version="0.0", insecure=False)
+        self.debug = debug
+        self._client = substra.Client(debug=debug, url=address, version="0.0", insecure=False)
         self._client.login(user, password)
 
     def add_data_sample(self, spec, *args, **kwargs):
--- a/substratest/factory.py
+++ b/substratest/factory.py
@@ -29,6 +29,7 @@ class TestOpener(tools.Opener):
                 reader = csv.reader(f)
                 for row in reader:
                     res.append(int(row[0]))
+        print(f'get_X: {{res}}')
         return res  # returns a list of 1's
     def get_y(self, folders):
         res = []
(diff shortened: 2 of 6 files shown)


substratest/factory.py
```diff
@@ -14,7 +14,7 @@
 
 DEFAULT_DATA_SAMPLE_FILENAME = 'data.csv'
 
-DEFAULT_SUBSTRATOOLS_VERSION = '0.6.0'
+DEFAULT_SUBSTRATOOLS_VERSION = '0.6.1'
```
AurelienGasser, line 17:
```suggestion
DEFAULT_SUBSTRATOOLS_VERSION = '0.6.1-minimal'
```

@Esadruhn I merged https://github.com/SubstraFoundation/substra-tests/pull/118 to use the minimal docker image, so you should be good to use the minimal version here too.

The `0.6.1-minimal` version was automatically built [on docker hub](https://hub.docker.com/layers/substrafoundation/substra-tools/0.6.1-minimal/images/sha256-2d061db2938587a05df3a4f66074d03f44f4fc74851a7f1d0ff645671ca0fbe8?context=repo)

tests/test_execution.py
```diff
@@ -132,13 +132,13 @@ def test_traintuple_execution_failure(factory, client, default_dataset_1):
         dataset=default_dataset_1,
         data_samples=default_dataset_1.train_data_sample_keys,
     )
-    if client.backend == "remote":
+    if client.debug:
```
samlesu, line 135: I think we should not have in the test different behaviour for debug and not debug mode.

I would prefer to flag these tests as remote only. 

As this change hasn't been introduced in this PR but we could handle that comment in a separate PR later.

(same comment for next tests)

substratest/client.py
```diff
@@ -14,13 +14,15 @@ class Client:
...
 
-    def __init__(self, backend, node_id=None, address=None, user=None, password=None):
+    def __init__(self, debug, node_id=None, address=None, token=None, user=None, password=None):
         super().__init__()
 
         self.node_id = node_id
-        self.backend = backend
-        self._client = substra.Client(backend=backend, url=address, version="0.0", insecure=False)
-        self._client.login(user, password)
+        self.debug = debug
```
samlesu, line 21: IMO we should not need to expose a debug attribute.

substratest/client.py
```diff
@@ -14,13 +14,15 @@ class Client:
...
+        self._client = substra.Client(debug=debug, url=address, version="0.0", insecure=False, token=token)
+        if not token:
+            self._client.login(user, password)
+        self.token = self._client._token
```
samlesu, line 25: We should avoid accessing a private attribute.

If I remember well the token is outputted by the login function. We could maybe use the following code instead?
```python
if not token:
    token = self._client.login()
self.token = token
```

substratest/factory.py
```diff
@@ -451,7 +479,13 @@ def add_composite_traintuple(self, composite_algo, dataset=None, data_samples=No
...
         return spec
 
-    def add_testtuple(self, objective, traintuple_spec, tag='', metadata=None):
+    def add_testtuple(
+        self,
+        objective,
+        traintuple_spec,
+        tag='',
+        metadata=None,
+    ):
```
samlesu, line 488: To ease the review, we should try to keep the code change as light as possible. 
So I think we should avoid unrelated code style change.

tests/conftest.py
```diff
@@ -115,7 +115,7 @@ def factory(request):
 def network(backend):
     """Network fixture.
 
-    Network must started outside of the tests environment and the network is kept
+    Network must ge started outside of the tests environment and the network is kept
```
samlesu, line 118:
```suggestion
    Network must be started outside of the tests environment and the network is kept
```

tests/conftest.py
```diff
@@ -115,7 +115,7 @@ def factory(request):
 def network(backend):
```
samlesu, line 115: I think we should get rid of the backend name here as it is not exposed in the client. Could it be converted to a boolean with an explicit name (for instance `client_debug` or `client_debug_enabled`) ?

This would allow to remove the TODO line 131.

tests/conftest.py
```diff
@@ -277,3 +277,17 @@ def client(network):
 def clients(network):
     """Client fixture (first node)."""
     return network.clients
+
+
+@pytest.fixture(scope="session")
+def debug_client(client):
```
samlesu, line 283: We could add a docstring to be consistent with other fixtures. For instance:
```python
"""Client fixture in debug mode (first node)."""
```

tests/test_debug.py
```diff
@@ -0,0 +1,108 @@
...
+        traintuple=traintuple,
+        data_samples=default_dataset.test_data_sample_keys[0],
+    )
+    testtuple = debug_client.add_testtuple(spec).future().wait()
+    assert testtuple.status == assets.Status.done
+    assert testtuple.dataset.perf == 3
+
+
+@pytest.mark.remote_only
+@pytest.mark.slow
```
samlesu, line 37: The test must be quite fast to run as the execution is local? (same for other tests)

Esadruhn: As a reference, `test_compute_plan_single_client_success` and `test_compute_plan_update ` take 7.55s to run in "local" mode in the CI. I guess we can expect the tests in test_debug to be of same or longer duration.
I don't have a clear grasp of `slow` in substra tests, what would be the rule of thumb ?

substratest/client.py
Esadruhn: Good catch !

tests/test_execution.py
Esadruhn: Yes, I think we could modify or move the 'fully local' tests entirely, it is better to do it in another PR as this won't be a single small change.

substratest/factory.py
Esadruhn: You are right, I try to avoid it, this one slipped through ! I revert it

tests/conftest.py
Esadruhn: Maybe `client_debug_local` ?  As this relates to the 'no Substra platform deployment' mode.

samlesu: Seems good to me! 👍

tests/test_debug.py
samlesu: Ok, I thought it would be much faster.
I don't think we have a proper rules for tagging a test as slow. The main idea is to ensure the "fast mode" can be used to run the tests for a quick development rules. So tagging a test as slow is a trade-off between use case coverage and performance.

Let's keep the slow tag for these tests then.

PR #113, opened 2020-06-29
Fix concurrency

AurelienGasser:
Companion PR:

- https://github.com/SubstraFoundation/substra-backend/pull/279

PR #109, opened 2020-06-04
Update values for k8s-tasks

Kelvin-M:
Companion PR : https://github.com/SubstraFoundation/substra-backend/pull/249

PR #126, opened 2020-07-28
K8s backend

Kelvin-M:
Non-mellody rebased branch
Include :
- k8s-tasks
- Fix concurrency

Companion PR:
SubstraFoundation/substra-backend#299
https://github.com/SubstraFoundation/hlf-k8s/pull/86

Related
https://github.com/SubstraFoundation/substra-tests/pull/109
https://github.com/SubstraFoundation/substra-tests/pull/113


values.yaml
```diff
@@ -7,10 +7,8 @@ nodes:
     address: 'http://substra-backend.node-1.com'
     user: 'node-1'
     password: 'p@$swr0d44'
-    shared_path: '/tmp/org-1/servermedias'
   - name: 'node-2'
     msp_id: 'MyOrg2MSP'
     address: 'http://substra-backend.node-2.com'
     user: 'node-2'
     password: 'p@$swr0d45'
-    shared_path: '/tmp/org-2/servermedias'
```
jmorel, line 16: Why remove the shared path? Is the feature of using shared path when adding data samples deprecated?

Kelvin-M: I removed sharedPath as server medias is not mount as a hostPath pv by default. With k8s-backend, pv use now the default storage class and not hostPath.
Adding data samples using a shared path between the host and pods is not deprecated :) I just miss to add an extra value to configure it (see https://github.com/SubstraFoundation/substra-backend/pull/304) !

jmorel: The trouble is that removing this shared_path value prevents `test_add_data_sample_located_in_shared_path` from running at all (it won't be executed). Is there a way we could still run this test with the new setup?

jmorel:
I'm sorry, I forgot to mention it in my previous comment : docker-compose is mentioned a few times in the README, could you edit it as well?

Kelvin-M:
Shared path volume is not symmetric anymore ! @jmorel 
shared_path value (host) is mounted at  `/var/substra/servermedias` on pods

ci/run-ci.py
```diff
@@ -161,7 +180,9 @@ def get_kube_context():
     global KUBE_CONTEXT
 
     print('\n# Fetch kubernetes context')
+    old_ctx = call_output('kubectl config current-context')
     call(f'gcloud container clusters get-credentials {CLUSTER_NAME} --zone {CLUSTER_ZONE} --project {CLUSTER_PROJECT}')
+    call(f'kubectl config use-context {old_ctx}') # Restore old context
```
jmorel, line 185: Nice!

values.yaml
```diff
@@ -8,9 +8,12 @@ nodes:
     user: 'node-1'
     password: 'p@$swr0d44'
     shared_path: '/tmp/org-1/servermedias'
+    # minikube: True
```
jmorel, line 11: Maybe you could move the minikube option to the `options` section? Since the value is currently only checked for the first node, it behaves like a general setup option and not something that may change on a node-to-node basis.

Kelvin-M: Yes, you're right !

substratest/factory.py
```diff
@@ -221,11 +222,24 @@ class DataSampleSpec(_Spec):
...
+            shutil.move(self.path, destination)
+        else:
+            destination = os.path.join(copy_destination, random_uuid()[0:8])
+
+            minikube_private_key = '~/.minikube/machines/minikube/id_rsa'
+            minikube_ssh = 'docker@$(minikube ip)'
+
+            os.system(f'scp -r -i {minikube_private_key} {self.path} {minikube_ssh}:/tmp/')
+            os.system(f'ssh -i {minikube_private_key} {minikube_ssh} '
+                      f'"sudo mkdir -p {destination} && sudo mv /tmp/{os.path.basename(self.path)} {destination}"')
```
samlesu, line 239: Could we do that in a single scp command?

If I understand well the ssh is required for the directory creation. 

Could we use instead an scp option (it looks like the `-r` option suits our need: `-r      Recursively copy entire directories.  Note that scp follows symbolic links encountered in the tree traversal.`) ?

substratest/factory.py
```diff
@@ -221,11 +222,24 @@ class DataSampleSpec(_Spec):
...
+
```
samlesu, line 240: To have the same behaviour with the not minikube case, we should maybe delete the orignal file after the scp?

substratest/factory.py
```diff
@@ -221,11 +222,24 @@ class DataSampleSpec(_Spec):
...
+        self.path = os.path.join(destination.replace(copy_destination, '/var/substra/servermedias/'),
```
samlesu, line 241: We could define `'/var/substra/servermedias/'` as a global constant.
For instance SERVER_MEDIA_PATH?

substratest/factory.py
```diff
@@ -221,11 +222,24 @@ class DataSampleSpec(_Spec):
     test_only: bool
     data_manager_keys: typing.List[str]
 
-    def move_data_to(self, destination):
-        destination = destination if destination.endswith('/') else destination + '/'
-        destination = tempfile.mkdtemp(dir=destination)
-        shutil.move(self.path, destination)
-        self.path = os.path.join(destination, os.path.basename(self.path))
+    def move_data_to(self, copy_destination, minikube=False):
```
samlesu, line 225: Why did you rename `destination` to `copy_destination`? I think `copy_` is misleading (the function is called `move_` and redundant with the function name.
A better name would be `destination_folder`? It would add more information to the code reader.

What do you think?

samlesu: We should maybe rename this function to `move_data_to_server()` as it is less generic than before. And it would help to understand the meaning of the minikube option.

Kelvin-M: Yes, good points !

substratest/factory.py
Kelvin-M: 👍

substratest/factory.py
Kelvin-M: The `-r` option is already used (l.237)
With scp or ssh with private key, we use only an user account and we need `root` permissions to write data in `{destination}` path (hostpath volume in minikube is under root user)

samlesu: Good point, thanks for the explanation. 

What do you think of doing first the mkdir, then a chown operation, and to finish the scp to the final destination? 

It would avoid the `mv` operation.

Kelvin-M: I tried to do that, but hostPath need to be root for k8s pods in minikube if we don't want to have permission issues inside :)

samlesu: Ok, thanks :)

substratest/factory.py
```diff
@@ -214,18 +215,34 @@ class PrivatePermissions(pydantic.BaseModel):
 
 DEFAULT_PERMISSIONS = Permissions(public=True, authorized_ids=[])
 DEFAULT_OUT_TRUNK_MODEL_PERMISSIONS = PrivatePermissions(authorized_ids=[])
-
+SERVER_MEDIA_PATH='/var/substra/servermedias/'
```
samlesu, line 218:
```suggestion
SERVER_MEDIA_PATH = '/var/substra/servermedias/'
```
make flake8 happy

PR #127, opened 2020-08-06
CI: Don't retry run-ci.py

AurelienGasser:
We currently retry run-ci.py in case it fails the first time. This strategy was put in place because the deployment used to randomly fail (PVC not bound). 

However:
- we haven't seen this PVC issue in a while, so we can presume it's fixed
- the newest change to substra-backend (k8s tasks) significantly increase the run time of run-ci.py. The time limit for a Travis job is 50 min﻿, which is not enough time for two full runs of run-ci.py. As a result, if run-ci.py fails the first time then`run-ci.py` is retried, but we hit the 50 min mark and the build is marked as 'errored'.

This PR removes the retry so that if run-ci.py fails, the build is marked as failed without retrying.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,8 +30,7 @@ jobs:
       script:
         - CLUSTER_NAME="substra-tests-$(date -u +'%Y-%m-%d-%Hh%M')"
         - cd ci/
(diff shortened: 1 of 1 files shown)


samlesu:
Thanks for the detailed PR description, it really helps 💯

PR #130, opened 2020-08-14
fix: expose docker socket to pod

Esadruhn:
Chose to expose the docker socket to the pod.

Source: article on Medium https://medium.com/hootsuite-engineering/building-docker-images-inside-kubernetes-42c6af855f25

They recommend the 'Docker-in-Docker' solution, however when going to the official docker hub https://hub.docker.com/_/docker, they recommend this blog post: https://jpetazzo.github.io/2015/09/03/do-not-use-docker-in-docker-for-ci/ that basically recommends the 'exposing socket' solution so I went with it, as it does not seem so bad and is certainly simpler.

I cannot test the solution for the moment, need either to launch the job manually from the CI or get the credentials to run `ci/run_ci.py` locally.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ pyclean:
 test: test-remote test-local
 
 test-remote: pyclean
-	pytest tests/test_debug.py -rs -v --durations=0 -n $(PARALLELISM)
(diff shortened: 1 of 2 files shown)


Esadruhn:
Closed in favor of #132

PR #132, opened 2020-08-14
fix: Docker in Docker solution to use Docker in the debug mode

Esadruhn:
Exposing the docker socket to the pod does not work because then we do not have permission to mount the volumes in RW on the Docker container

So this is the Docker in Docker method

## Old references

Source: article on Medium https://medium.com/hootsuite-engineering/building-docker-images-inside-kubernetes-42c6af855f25

They recommend the 'Docker-in-Docker' solution, however when going to the official docker hub https://hub.docker.com/_/docker, they recommend this blog post: https://jpetazzo.github.io/2015/09/03/do-not-use-docker-in-docker-for-ci/ that basically recommends the 'exposing socket' solution so I went with it, as it does not seem so bad and is certainly simpler.

I cannot test the solution for the moment, need either to launch the job manually from the CI or get the credentials to run ci/run_ci.py locally.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ pyclean:
 test: test-remote test-local
 
 test-remote: pyclean
-	pytest tests/test_debug.py -rs -v --durations=0 -n $(PARALLELISM)
+	pytest tests -rs -v --durations=0 -n $(PARALLELISM)
 
 test-minimal: pyclean
 	pytest tests -rs -v --durations=0 -m "not slow" -n $(PARALLELISM)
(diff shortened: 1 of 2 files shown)


AurelienGasser:
@Kelvin-M FYI we tried mounting the docker socket but then the application (substra) tries to `docker run -v /usr/src/app/foo:/usr/src/app/foo` which doesn't work because `/usr/src/app/foo` here refers to the docker host (the GKE machine) and not the pod FS.

AurelienGasser:
Note: E2e tests don't pass on GKE as of right now: there's another independent issue in the tests.

PR #134, opened 2020-08-17
Fix the tests in test_debug.py

Esadruhn:
The 'dind' (Docker in Docker) container and the 'substra-tests' container must share the '/tmp' volume to be able to mount it in the Docker containers spawned on the fly for the traintuple execution.  
Same for the 'local-worker' folder

### DIFF
--- a/tests/test_debug.py
+++ b/tests/test_debug.py
@@ -4,7 +4,7 @@ from substratest import assets
 from substratest.factory import Permissions
 
 
-@pytest.mark.skip("Need to fix this test")
+# @pytest.mark.skip("Need to fix this test")
 @pytest.mark.remote_only
 @pytest.mark.slow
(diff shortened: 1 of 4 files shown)


tests/test_debug.py
```diff
@@ -4,7 +4,7 @@
 from substratest.factory import Permissions
 
 
-@pytest.mark.skip("Need to fix this test")
+# @pytest.mark.skip("Need to fix this test")
```
Kelvin-M, line 7: @Esadruhn This comment looks like a leftover now, no?

Esadruhn: Yes, we can remove it in the next PR ? Or I can do a small one to remove it now

Kelvin-M: As you want !

PR #138, opened 2020-08-27
CI: Fix substra version when using cache

Esadruhn:
When we merge a PR in substra and its companion PR in substra-tests the same day, the latest changes in substra are not taken into account and the tests fail.

This is because the CI uses a Kaniko cache (168h) for the substra-tests image.

**Changes**: 
- use correct substra version when running the CI with cache
- also remove useless commented lines

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,7 +43,7 @@ jobs:
     - name: "Tests on the local backend"
       env:
         - SUBSTRA_GIT_REPO=https://github.com/SubstraFoundation/substra.git
-        - SUBSTRA_GIT_REF=master
+        - SUBSTRA_BRANCH=master
       before_install:
         # Update Docker to the right version
         - sudo apt-get update
@@ -53,7 +53,7 @@ jobs:
         - sudo apt-get install docker-ce=5:19.03.12~3-0~ubuntu-xenial docker-ce-cli=5:19.03.12~3-0~ubuntu-xenial containerd.io=1.2.13-2
         - docker version
         # Install substra and substra-tests (for the tests on the local backend)
-        - pip install --no-cache-dir "git+${SUBSTRA_GIT_REPO}@${SUBSTRA_GIT_REF}"
+        - pip install --no-cache-dir "git+${SUBSTRA_GIT_REPO}@${SUBSTRA_BRANCH}"
         - pip install --no-cache-dir -r requirements.txt
       script:
         - make test-local
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,7 +39,7 @@ jobs:
       script:
         - CLUSTER_NAME="substra-tests-$(date -u +'%Y-%m-%d-%Hh%M')"
         - cd ci/
-        - python -u ./run-ci.py --keys-directory=./keys/ --cluster-name=${CLUSTER_NAME}
+        - python -u ./run-ci.py --no-cache --keys-directory=./keys/ --cluster-name=${CLUSTER_NAME}
     - name: "Tests on the local backend"
       env:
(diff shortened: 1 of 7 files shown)


.travis.yml
```diff
@@ -39,7 +39,7 @@ jobs:
       script:
         - CLUSTER_NAME="substra-tests-$(date -u +'%Y-%m-%d-%Hh%M')"
         - cd ci/
-        - python -u ./run-ci.py --keys-directory=./keys/ --cluster-name=${CLUSTER_NAME}
+        - python -u ./run-ci.py --no-cache --keys-directory=./keys/ --cluster-name=${CLUSTER_NAME}
```
Kelvin-M, line 42: @AurelienGasser what do you think about that ?

AurelienGasser: I had a quick look at this. 

This PR completely disables the cache for the CI, I think we can avoid doing that.

The root of the issue seems to be that the output image tag `substra-tests:ci-xxx` only depends on the SHA of the `substra-tests` commit, not the `substra` commit.

https://github.com/SubstraFoundation/substra-tests/blob/9e32fdf01c750dc67c1cf302b6d1e2bbfed3d4f9/ci/cloudbuild/substra-tests.yaml#L6

^ here `_COMMIT` refers to substra-tests.

A potential fix:

1. In `run-ci.py`, stop supporting any "ref" for substra, and only support branch names, like for the other projects (see [here](https://github.com/SubstraFoundation/substra-tests/blob/master/ci/run-ci.py#L60-L63). In other words, rename `SUBSTRA_GIT_REF` to `SUBSTRA_BRANCH`.
2. In `run-ci.py`, fetch the latest commit corresponding to the substra branch (see [here](https://github.com/SubstraFoundation/substra-tests/blob/master/ci/run-ci.py#L287-L293))
3. Pass it to `substra-tests.yaml` as a new substition parameter (see [here](https://github.com/SubstraFoundation/substra-tests/blob/master/ci/run-ci.py#L346))
4. In `substra-tests.yaml`, use the commit SHA in the image tag

```diff
- --destination=eu.gcr.io/$PROJECT_ID/substra-tests:ci-${_COMMIT}
+ --destination=eu.gcr.io/$PROJECT_ID/substra-tests:ci-${_COMMIT}-${_SUBSTRA_GIT_COMMIT}
```

Esadruhn: Thanks for the detailed answer ! I changed it so that the image depends on the commit of substra and substra-tests, it seems to work well

skaffold.yaml
```diff
@@ -22,7 +22,7 @@ build:
         dockerfile: docker/substra-tests/Dockerfile
         buildArgs:
           SUBSTRA_GIT_REPO: https://github.com/SubstraFoundation/substra.git
-          SUBSTRA_GIT_REF: master
+          SUBSTRA_GIT_COMMIT: master
```
AurelienGasser, line 25: looks like there's a discrepancy here: the variable name is `COMMIT` but the value is `master`

Esadruhn: Then I should keep the 'git_ref' name in the Dockerfile since it does `pip install git+{URL}@{REF}` where REF can either be a commit or branch (but `HEAD` does not work)

AurelienGasser: Yep that sounds good!

AurelienGasser:
@Esadruhn please let me know when this is ready for review, I'll give it a test run

Esadruhn:
Yes, ready for review !

AurelienGasser:
@Esadruhn 

- I update this PR's title/description
- I added another PR https://github.com/SubstraFoundation/substra-tests/pull/139#issue-476437171 with an additional fix + minor improvement.

PR #142, opened 2020-09-09
Switch from camel case to snake case

Kelvin-M:
The backend responses should be more consistent. The goal of this PR is to make sure that all inputs fields and response fields of backend endpoints are in snake_case format.

PRs:
https://github.com/SubstraFoundation/substra-backend/pull/315
https://github.com/SubstraFoundation/substra/pull/222
https://github.com/SubstraFoundation/substra-tests/pull/142

### DIFF
--- a/substratest/assets.py
+++ b/substratest/assets.py
@@ -129,22 +129,6 @@ class _ComputePlanFutureMixin(_BaseFutureMixin):
     _future_cls = ComputePlanFuture
 
 
-_MANUAL_CONVERTION_FIELDS = {
-    'data_manager_key': 'dataset_key',
-}
-
-
-def convert_asset_field_names(name):
-    """Converts asset camel case fields to snake case fields.
-
-    Special cases are handled through a global dict variable.
-    """
-    # XXX using a mapper for converting specific is not very flexible as it will be
-    #     applied to all fields from all assets.
-    mapper = _MANUAL_CONVERTION_FIELDS
-    return mapper[name] if name in mapper else name
-
-
 class _InternalStruct(pydantic.BaseModel, abc.ABC):
     """Internal nested structure."""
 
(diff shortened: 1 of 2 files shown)


substratest/assets.py
```diff
@@ -130,8 +130,7 @@ class _ComputePlanFutureMixin(_BaseFutureMixin):
 
 
 _MANUAL_CONVERTION_FIELDS = {
-    'authorizedIDs': 'authorized_ids',
-    'dataManagerKey': 'dataset_key',
+    'data_manager_key': 'dataset_key',
```
Kelvin-M, line 133: @jmorel Do you remember why we convert data manager to dataset ?

Esadruhn:
There are some values in the datastore in CamelCase

Kelvin-M:
> There are some values in the datastore in CamelCase, I don't know why it doesn't make tests fail.
> 
> eg
> 
> * TESTTUPLE, line 75 'traintupleKey'
> * 'isCurrent'
> 
> will test it when I can (in a meeting)

I fixed it !

Esadruhn (dismissed):

substratest/utils.py
```diff
@@ -24,12 +23,6 @@ def create_archive(tmpdir, *files):
     return zip_folder(str(tmpdir))
 
 
-def camel_to_snake(name):
-    """Convert camel case to snake case."""
-    s = re.sub('(.)([A-Z][a-z]+)', r'\1_\2', name)
-    return re.sub('([a-z0-9])([A-Z])', r'\1_\2', s).lower()
-
-
 def replace_dict_keys(d, converter):
```
jmorel, line 26: Since this method is now only used to have a dataset_key on the class ObjectiveDataset, maybe we could remove it entirelyn, update the ObjectiveDataset class to have a datamanager_key property and add a dataset_key @property that would return the datamanager_key value. What do you think?

Kelvin-M: Good Idea, can you do the update or do you want me to do it ?

substratest/assets.py
jmorel: I thought I already posted a reply to this, but it's not in this PR, so here it is again :)

The idea is that data_manager is an old name that we try to phase out in favor of dataset. This allows the tests developers to not think about the old name and only think in terms of dataset.

substratest/utils.py
jmorel: On it :)

jmorel: So it looks like the dataset_key property is not used anywhere, so I just renamed it in data_manager_key and deleted the related code. Could you run the full test suite to make sure it's still ok?

jmorel: The test_network module still passes btw, so I'm fairly confident.

jmorel (approved):

PR #144, opened 2020-09-14
Susbtra returns an object instead of a dict

Esadruhn:
This is the companion PR to: https://github.com/SubstraFoundation/substra/pull/225


AurelienGasser (dismissed):
Great work!

substratest/client.py
```diff
@@ -201,15 +150,44 @@ def describe_dataset(self, key):
...
         self._client.link_dataset_with_objective(dataset.key, objective.key)
         # XXX do not return anything as currently the chaincode simply returns the
         #     updated dataset key
 
     def link_dataset_with_data_samples(self, dataset, data_samples):
-        data_sample_keys = [d.key for d in data_samples]
-        self._client.link_dataset_with_data_samples(dataset.key, data_sample_keys)
+        self._client.link_dataset_with_data_samples(dataset.key, data_samples)
+
+    def get_compute_plan_traintuple(self, compute_plan_id):
```
jmorel, line 163: This should be named `list_compute_plan_traintuples`

jmorel: Also applies to the 3 other get_compute_plan_*tuple methods :)

tests/test_data_samples_order.py
```diff
@@ -77,9 +78,9 @@ def test_traintuple_data_samples_relative_order(factory, client, default_dataset
     traintuple = client.add_traintuple(spec)
     # Ensure the order of the data sample keys is correct at 2 levels: :
     #  1. In the returned traintuple
-    #  2. In the train method of the algo. If the order is incorrect, .future().wait() will fail.
+    #  2. In the train method of the algo. If the order is incorrect, Future(...).wait() will fail.
     assert traintuple.dataset.keys == data_sample_keys
-    traintuple.future().wait()
+    assets.Future(traintuple, client).wait()
```
jmorel, line 83: What do you think of wrapping the sdk response in Future not in the tests but rather in the Client directly?

tests/test_network.py
```diff
@@ -99,7 +99,7 @@ def test_link_dataset_with_datasamples(factory, client):
     client.link_dataset_with_data_samples(dataset, [data_sample])
 
     dataset = client.get_dataset(dataset.key)
-    assert dataset.train_data_sample_keys == [data_sample.key]
+    assert dataset.train_data_sample_keys == [data_sample]
```
jmorel, line 102: it looks like the `data_sample` variable should really be named `data_sample_key`.

tests/test_permissions.py
```diff
@@ -124,8 +126,9 @@ def test_merge_permissions(permissions_1, permissions_2, expected_permissions,
         dataset=dataset_1,
         data_samples=[train_data_sample_1],
     )
-    traintuple = client_1.add_traintuple(spec).future().wait()
-    assert traintuple.status == assets.Status.done
+    traintuple = client_1.add_traintuple(spec)
+    traintuple = assets.Future(traintuple, client_1).wait()
+    assert traintuple.status == models.Status.done
```
jmorel, line 131: the wait method used to raise an error if the status wasn't done, I'm not sure if the behavior has been changed in this PR, but this is the reason why there was no assertion on status.

tests/test_data_samples_order.py
jmorel: In retrospect, I'm not sure that wrapping a whole class is such a good idea. But maybe we could replace the class with a simpler function call ? `wait(traintuple, client)` or maybe `client.wait(traintuple)` ?

This doesn't change the fact that it's otherwise working fine!

jmorel (changes requested):

tests/test_permissions.py
Esadruhn: There was one here, you are right it is unnecessary

tests/test_data_samples_order.py
Esadruhn: True, it can now be simplified to `client.wait`, no need for the whole `Future` class. I tried to wrap it to see how it was, did not see a clean way of doing it.  
Seeing how it was used in the code, I thought that this calling `sthg(my_asset).wait()` did not change the usage too much.

jmorel (approved):
I couldn't test this fully but I like what I see! Very nice job!

PR #152, opened 2020-10-01
CI: Various minor improvements

AurelienGasser:
- bump kubernetes version
- changed the default susbtra-tests timeout from 300 (default) to 400 to mitigate "future timeout" errors
- added --machine-type param
- remove warnings at GKE cluster creation
- rename a few params/variables for clarity
﻿

PR #154, opened 2020-10-26
Add fetch cluster version function

Kelvin-M:
GKE regularly changes cluster version list compatible as master node.
In this PR, we directly fetch the cluster version from `REGULAR` channel in order to avoid failure from deprecated version value.
https://cloud.google.com/kubernetes-engine/docs/release-notes

### DIFF
--- a/ci/run-ci.py
+++ b/ci/run-ci.py
@@ -109,23 +109,6 @@ def cluster_name(value):
 
     return value
 
-
-def cluster_version():
-    """
-    Fetch gcloud last regular version
-    This is to ensure the cluster is up to date with the regular channel version.
-    """
-    cluster_versions = call_output(f'gcloud container get-server-config --zone={CLUSTER_ZONE} --format=json')
-    # Remove first line because it's a log info from gcloud cli
-    cluster_versions = '\n'.join(cluster_versions.split('\n')[1:])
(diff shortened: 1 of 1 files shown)


ci/run-ci.py
```diff
@@ -213,7 +231,7 @@ def get_kube_context():
 def create_cluster_async():
     print('\n# Create GKE cluster')
     cmd = f'gcloud container clusters create {CLUSTER_NAME} '\
-          f'--cluster-version {CLUSTER_VERSION} '\
+          f'--cluster-version {cluster_version()} '\
```
AurelienGasser, line 234: Maybe we could simply not specify the `--cluster-version` parameter and let GKE use its default version? I just tested it and it works.

Kelvin-M: Yes. Nevertheless, the default one for GKE is stable, which is 1.16.x. It's different from the bump you've done here : https://github.com/SubstraFoundation/substra-tests/pull/152

AurelienGasser: I think it's ok if we stay on stable. Actually it's probably better to use stable, what do you think?

AurelienGasser:
tested: :heavy_check_mark:

AurelienGasser (approved):
Nice!

PR #156, opened 2020-10-28
ci: add chaincode branch option + fix concurrency override

AurelienGasser:


Kelvin-M (approved):
Nice job ! 
The chaincode part won't work in the future with hlf-2 as it will be a docker image.

PR #158, opened 2020-11-03
Rename a few attributes

AurelienGasser:
Companion PRs:

- https://github.com/SubstraFoundation/substra-backend/pull/345
- https://github.com/SubstraFoundation/substra-chaincode/pull/129
- https://github.com/SubstraFoundation/substra/pull/241

Rename:

| Before | After |
|---|----|
| `compute_plan_id` | `compute_plan_key` (or simply `key`) |
|  `hash` | `checksum` |
| `data_manager.keys` | `data_manager.data_sample_keys` |

### DIFF
--- a/substratest/client.py
+++ b/substratest/client.py
@@ -209,10 +209,7 @@ class Client:
             assert False, 'Future not supported'
         getter = getattr(self, m)
 
-        if asset.__class__.__name__ == "ComputePlan":
-            key = asset.compute_plan_key
-        else:
-            key = asset.key
+        key = asset.key
 
         tstart = time.time()
         while asset.status not in [Status.done.value, Status.failed.value, Status.canceled.value]:
--- a/substratest/client.py
+++ b/substratest/client.py
@@ -80,8 +80,8 @@ class Client:
     def update_compute_plan(self, spec, *args, **kwargs):
         spec_dict = spec.dict()
         # Remove extra field from data
-        spec_dict.pop("compute_plan_key")
-        return self._client.update_compute_plan(spec.compute_plan_key, spec_dict, *args, **kwargs)
+        spec_dict.pop("key")
+        return self._client.update_compute_plan(spec.key, spec_dict, *args, **kwargs)
 
     def list_compute_plan(self, *args, **kwargs):
         return self._client.list_compute_plan(*args, **kwargs)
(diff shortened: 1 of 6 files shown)


jmorel (approved):

substratest/client.py
```diff
@@ -209,10 +209,7 @@ def wait(self, asset, timeout=cfg.FUTURE_TIMEOUT, raises=True):
             assert False, 'Future not supported'
         getter = getattr(self, m)
 
-        if asset.__class__.__name__ == "ComputePlan":
-            key = asset.compute_plan_id
-        else:
-            key = asset.key
+        key = asset.key
```
jmorel, line 212: Can we just remove this variable and use asset.key everywhere?

jmorel (changes requested):

substratest/client.py
```diff
@@ -80,8 +80,8 @@ def add_compute_plan(self, spec, *args, **kwargs):
     def update_compute_plan(self, spec, *args, **kwargs):
         spec_dict = spec.dict()
         # Remove extra field from data
-        spec_dict.pop("compute_plan_id")
-        return self._client.update_compute_plan(spec.compute_plan_id, spec_dict, *args, **kwargs)
+        spec_dict.pop("key")
+        return self._client.update_compute_plan(spec.key, spec_dict, *args, **kwargs)
```
Kelvin-M, line 84: Why did you rename it as `key` and not `compute_plan_key` ? 
This modification occurs multiple time in this PR :)

Kelvin-M (changes requested):

substratest/client.py
AurelienGasser: Comment addressed in https://github.com/SubstraFoundation/substra/pull/241#issuecomment-721759361

substratest/client.py
AurelienGasser: The only place where we use this variable is 

```
asset = getter(key)
```

I think changing it to 

```
asset = getter(asset.key)
```

might potentially bring new categories of bugs: if `getter` doesn't return the correct result, then the subsequent `getter` call will use the wrong argument. For this reason I think it's better to create a dedicated variable `key` to ensure `key` is constant throughout the execution of this function.

What do you think?

jmorel: Makes sense, thank you for the explanation!

jmorel (approved):

AurelienGasser:
The travis-ci tests won't pass until the substra PR is merged.

Kelvin-M (approved):

PR #168, opened 2020-12-03
ci: add hacky support for tag parameters

AurelienGasser:
Currently, the CI script only support branches as arguments.

This PR prevents the script from crashing when trying to get the digest of a git tag, with commands like:

```
run-ci.py --substra-tests 0.6.0
```

### DIFF
--- a/ci/run-ci.py
+++ b/ci/run-ci.py
@@ -353,7 +353,12 @@ def clone_repository(dirname, url, branch, commit=None):
     call(f'git clone -q --depth 1 {url} --branch "{branch}" {dirname}')
 
     if commit is None:
-        commit = call_output(f'git --git-dir={dirname}/.git rev-parse origin/{branch}')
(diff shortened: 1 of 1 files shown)


Kelvin-M:
You should update argpase helper and run-ci.py too to be more explicit as `branch` var name is used everywhere :D

AurelienGasser:
@Kelvin-M I'd rather not complicate things too much and keep "branch" everywhere. I don't want to add full support for tags at this point, just to unblock @jmorel. I'll update the PR name/description accordingly.

Kelvin-M:
Is it still relevant ?

jmorel:
I don't have a need for it anymore myself, but maybe a check with other contributors would be a good idea.

Kelvin-M (approved):

Kelvin-M (approved):

PR #170, opened 2020-12-17
Migrate from Travis CI to GitHub Actions

AlexandrePicosson:
## Description

This PR adds a new GH Actions workflow to run the end to end tests it is the same workflow as the one that was run on TravisCI.

The GCP auth key is encrypted using gpg with AES256 as algorithm as described [here](https://docs.github.com/en/free-pro-team@latest/actions/reference/encrypted-secrets#limits-for-secrets) as we cannot rely on the travis secret solution we used before. 

The only breaking change is the removal of the slack notification until we find a solution to do this correctly with GitHub Actions. Another change that could have an impact is the fact that we rely on the runner docker version and we do not install a specific version, let me know if you would prefer having a fixed version as before.

When does the CI run ?
- **end to end on gcp**: Every day at 00:00
- **Local backend**: On every PR to master, every merge to master, every day at 00:00

## Motivation and Context

We switch from TravisCI to GH Actions due to the fact that https://travis-ci.org/ will close at the end of december ([source](https://docs.travis-ci.com/user/migrate/open-source-repository-migration#q-when-will-the-migration-from-travis-ciorg-to-travis-cicom-be-completed)).

## How Has This Been Tested?

I triggered manually the end to end gcp tests: [e2e pipeline result](https://github.com/SubstraFoundation/substra-tests/actions/runs/427896105).

You can look at the [Checks](https://github.com/SubstraFoundation/substra-tests/pull/170/checks) tab to see the workflow logs.

## Screenshots
Preview of the end to end gcp tests:
Screenshot 2020-12-17 at 13 14 42

## Types of changes

- Bug fix (non-breaking change which fixes an issue)
- New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.


.github/workflows/end-to-end.yml
```diff
@@ -0,0 +1,40 @@
...
+      - uses: actions/setup-python@v2
+        name: Set up python
+        with:
+          python-version: 3.7
+      - uses: azure/setup-helm@v1
+        with:
+            version: 'v3.2.3'
+      - uses: azure/setup-kubectl@v1
+        with:
+            version: 'latest'
```
inalgnu, line 23: shouldn't we have a fixed version here?

AlexandrePicosson: The previous script used latest but we can change this. I don't think there is an issue with using latest as it seems quite unlikely that k8s releases a version with a breaking change.

Kelvin-M (dismissed):
LGTM !

AurelienGasser:
Great work!

> The only breaking change is the removal of the slack notification

Is an alerting mechanism offered by GitHub Actions for when builds fail? IMO we should have some sort of mechanism to alert us of CI build failures, otherwise CI build failures will most likely go unnoticed. 

Again, nice work!

AlexandrePicosson:
I know you can opt in to receive emails for personal repositories or when one of your pull requests builds fails in your account settings. I'll check if there is a similar setting for organisation accounts / cron builds.

AlexandrePicosson:
Apparently activating this setting https://docs.github.com/en/free-pro-team@latest/github/managing-subscriptions-and-notifications-on-github/configuring-notifications#github-actions-notification-options should be enough to receive an email or  a web notification for a failed workflow (pr build or cron). Also we could use another action to send an email but we would need a mail server.

mblottiere:
LGTM 👍 
Regarding notifications, some [actions on the marketplace](https://github.com/marketplace/actions/slack-notify-build) provide a similar feature:
```yaml
- name: Notify slack fail
  if: failure()
  env:
    SLACK_BOT_TOKEN: ${{ secrets.SLACK_BOT_TOKEN }}
  uses: voxmedia/github-action-slack-notify-build@v1
  with:
    channel: app-alerts
    status: FAILED
    color: danger
```
However, this does not cover the "on change" trigger offered by travis. Still, it's maybe better than nothing?

AurelienGasser:
> However, this does not cover the "on change" trigger offered by travis

IMO having a Slack notification every 24h is fine.

Kelvin-M:
Where are we on this one, it seems important to switch from travis to github action !

mblottiere (approved):

.github/workflows/python.yml
```diff
@@ -27,12 +27,9 @@ jobs:
           pip install --no-cache-dir "git+${SUBSTRA_GIT_REPO}@${SUBSTRA_BRANCH}"
           pip install --no-cache-dir -r requirements.txt
       - name: Test
-        run: make test-local
-  notification:
-    needs: test
-    runs-on: ubuntu-latest
-    steps:
+        run: make test-loca
```
Kelvin-M: missing `l`

inalgnu (approved):

AurelienGasser (approved):
Nice work!

PR #181, opened 2021-03-26
fix: Update objective specification and remove shuffling in datasample order tests

Kelvin-M:
Fix issue in datasamples order tests
Update objective specification which need to have test-only datasamples
Update testtuple specification to be more specific on sample used (different of the train)

Companion PR https://github.com/SubstraFoundation/substra-chaincode/pull/140


Kelvin-M:
E2E tests pass

jmorel (approved):

PR #185, opened 2021-05-05
feat: update tests to handle metadata

Kelvin-M:
Companion PRs : https://github.com/SubstraFoundation/substra-tools/pull/46, https://github.com/SubstraFoundation/substra-backend/pull/403

For this PR, we will need to do a release on substra-tools (0.8.0) to make it work properly.

### DIFF
--- a/substratest/factory.py
+++ b/substratest/factory.py
@@ -16,7 +16,7 @@ from . import utils, assets
 
 DEFAULT_DATA_SAMPLE_FILENAME = 'data.csv'
 
-DEFAULT_SUBSTRATOOLS_VERSION = '0.7.0-minimal'
+DEFAULT_SUBSTRATOOLS_VERSION = '0.8.0-minimal'
 
 DEFAULT_OPENER_SCRIPT = f"""
 import csv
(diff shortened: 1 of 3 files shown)


AurelienGasser:
Nice work? What do you think of adding a backwards compatibility test that would ensure that algos work even if they don't define a `metadata` parameter to the train function?

Kelvin-M:
> Nice work! What do you think of adding a backwards compatibility test that would ensure that algos work even if they don't define a `metadata` parameter to the train function?

Kelvin-M:
> Nice work! What do you think of adding a backwards compatibility test that would ensure that algos work even if they don't define a `metadata` parameter to the train function?

It won't be backward compatible because the substra-backend will put the `-- metadata XXX` argument if there are metadata in the tuple, so argparse will raise a `error: unrecognized arguments: --metadata XXX`

AurelienGasser:
Here I'm talking about compatibility with existing algos, not between the backend and substra-tools

AurelienGasser:
@Kelvin-M IMO this is problematic. Isn't there a way to add optional arguments with argparse? The altternative is that every user needs to rewrite their algo classes, which isn't great

Kelvin-M:
There is a solution for that (see [here](https://stackoverflow.com/a/12818237)) for instance, but users will also need to rewrite their algo.
The issue does not come from the algo, but from substra-backend that cannot know if it should pass `--metadata` or not

PR #186, opened 2021-05-13
test: add local folder test

AurelienGasser:
Until now, we didn't have tests for the "local folder" feature﻿.

### DIFF
--- a/tests/test_execution_compute_plan.py
+++ b/tests/test_execution_compute_plan.py
@@ -599,7 +599,6 @@ if __name__ == '__main__':
 """
 
 @pytest.mark.slow
-@pytest.mark.remote_only
 def test_compute_plan_local_folder(factory, client, default_dataset, default_objective_1):
     data_sample_1, data_sample_2, _, _ = default_dataset.train_data_sample_keys
 
--- a/tests/test_execution_compute_plan.py
+++ b/tests/test_execution_compute_plan.py
@@ -8,7 +8,7 @@ from substratest import assets
 
 
 @pytest.mark.remote_only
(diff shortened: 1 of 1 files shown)


AurelienGasser:
e2e :heavy_check_mark:

Kelvin-M (approved):
LGTM

tests/test_execution_compute_plan.py
```diff
@@ -553,3 +553,83 @@ def test_compute_plan_no_batching(factory, client, default_dataset, default_obje
...
+    def save_model(self, model, path):
+        with open(path, 'w') as f:
+            return json.dump(model, f)
+
+if __name__ == '__main__':
+    tools.algo.execute(TestAlgo())
+"""
+
+@pytest.mark.slow
+@pytest.mark.remote_only
```
Esadruhn, line 602: This test also works in local mode, is there a reason to mark it as remote only ?

AurelienGasser: I marked it as remote only because the first test in this file (`test_compute_plan_simple`) has the same decorator. Maybe it could be removed from there too?

Esadruhn: `test_compute_plan_simple` is remote only because it relies on several clients. `test_compute_plan_no_batching` should work in local mode too

AurelienGasser: Gotcha. Fixed in https://github.com/SubstraFoundation/substra-tests/pull/186/commits/48025546fbe97e59ea72eede2891d46b73cd33b4

AurelienGasser:
@Kelvin-M Can you approve again please

Kelvin-M (approved):

PR #201, opened 2022-09-08
test: complete the test for transient outputs

AlexandrePicosson:
## Description

Test that a transient output effectively has no address after processing (meaning it was properly deleted) and test that we cannot create a new task that depend on this model as an input.

## Motivation and Context

Sometimes ago I introduced a very basic test for transient outputs that just checked if the flag was set. Since then I made some progress and completed the work on the backend side to disable transient outputs properly. It's now time to test the full pipeline.

This test is simpler than the old `test_compute_plan_remove_intermediary_model` because here we are only checking the generic use case: a task with a transient outputs and not a complete workflow with train/predic/test. 

## How Has This Been Tested?

Running the test against a locally deployed Substra stack.

### DIFF
--- a/substratest/client.py
+++ b/substratest/client.py
@@ -1,5 +1,6 @@
 import tempfile
 import time
+import typing
 from typing import Optional
 
 import requests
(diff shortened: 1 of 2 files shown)


AlexandrePicosson:
Note: for this to work and pass we need to merge https://github.com/Substra/substra-backend/pull/439

sergebouchut2 (approved):

PR #211, opened 2022-09-20
feat: remove Model.category

mblottiere:
This field is being deprecated, let's not use it.

Companion PR for https://github.com/Substra/substra/pull/288

### DIFF
--- a/tests/test_execution_compute_plan.py
+++ b/tests/test_execution_compute_plan.py
@@ -579,9 +579,8 @@ def test_compute_plan_aggregate_composite_traintuples(  # noqa: C901
     # Check that permissions were correctly set
     for task in composite_traintuples:
         task = clients[0].get_composite_traintuple(task.key)
-        trunks = [model for model in task.composite.models if model.category == models.ModelType.simple]
-        for trunk in trunks:
-            assert len(trunk.permissions.process.authorized_ids) == len(clients)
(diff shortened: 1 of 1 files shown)


thibowk (approved):

tests/test_execution_compute_plan.py
```diff
@@ -579,9 +579,8 @@ def test_compute_plan_aggregate_composite_traintuples(  # noqa: C901
     # Check that permissions were correctly set
     for task in composite_traintuples:
         task = clients[0].get_composite_traintuple(task.key)
-        trunks = [model for model in task.composite.models if model.category == models.ModelType.simple]
-        for trunk in trunks:
-            assert len(trunk.permissions.process.authorized_ids) == len(clients)
+        trunk = task.outputs[OutputIdentifiers.shared].value
```
sergebouchut2, line 582: [minor] Shouldn't we use another naming if truck is deprecated?
```suggestion
        shared_models = task.outputs[OutputIdentifiers.shared].value
```

sergebouchut2 (approved):
lgtm

tests/test_execution_compute_plan.py
mblottiere: Eventually yes, but since there are still plenty of trunks/heads in the test I'll leave it as is for now.
Tests will need a good overhaul once we are done with generic tasks, let's rename everything at that time.

PR #221, opened 2022-09-30
feat: use generic task

Esadruhn:
https://github.com/Substra/substra-backend/pull/476
https://github.com/Substra/substrafl/pull/31
https://github.com/Substra/substra/pull/299

### DIFF
--- a/substratest/factory.py
+++ b/substratest/factory.py
@@ -9,15 +9,10 @@ import uuid
 
 import pydantic
 import substra
-from substra.sdk.schemas import AggregatetupleSpec
 from substra.sdk.schemas import AlgoSpec
-from substra.sdk.schemas import CompositeTraintupleSpec
-from substra.sdk.schemas import ComputePlanPredicttupleSpec
-from substra.sdk.schemas import ComputePlanTesttupleSpec
+from substra.sdk.schemas import ComputePlanTaskSpec
 from substra.sdk.schemas import Permissions
-from substra.sdk.schemas import PredicttupleSpec
-from substra.sdk.schemas import TesttupleSpec
-from substra.sdk.schemas import TraintupleSpec
+from substra.sdk.schemas import TaskSpec
 from substra.sdk.schemas import UpdateAlgoSpec
 from substra.sdk.schemas import UpdateComputePlanSpec
 from substra.sdk.schemas import UpdateDatasetSpec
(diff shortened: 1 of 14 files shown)


Esadruhn:
`make test-subprocess` pass

sergebouchut2 (approved):
LGTM

substratest/factory.py
```diff
@@ -402,162 +397,86 @@ def __init__(self, dataset) -> None:
...
-
-    assert len(traintuple_ids) == 1
-    traintuple_id = traintuple_ids.pop()
-
-    return traintuple_id
-
-
 class _ComputePlanSpecFactory:
-    def create_traintuple(self, algo, inputs=None, outputs=None, tag="", metadata=None) -> ComputePlanTraintupleSpec:
+    def create_traintuple(self, algo, worker, inputs=None, outputs=None, tag="", metadata=None) -> ComputePlanTaskSpec:
```
sergebouchut2, line 401: Maybe these methods are not so much necessary anymore. We can define default outputs upstream and use a generic create_task method. But this could be easily done later.

Esadruhn: As you say, we should review our whole test process for generic task, some tests are now redundant
These methods are useful to have the right inputs/outputs for each task, and connect depending tasks correctly for now

tests/test_execution_compute_plan.py
```diff
@@ -493,109 +509,109 @@ def test_compute_plan_aggregate_composite_traintuples(  # noqa: C901
...
-                        and i.parent_task_key == task.in_head_model_id
+                        and i.parent_task_key
+                        == [x for x in task.inputs if x.identifier == InputIdentifiers.local][0].parent_task_key
                         and i.parent_task_output_identifier == InputIdentifiers.local
                     ]
                 )
                 == 1
             )
-        if task.in_trunk_model_id:
+        print(task.inputs)
```
louishulot, line 569: TODO: remove

louishulot (approved):
LGTM, print statement should be removed though

louishulot:
test are passing in local on my side

sergebouchut2 (changes requested):
[E2e tests fail](https://github.com/owkin/substra-ci/actions/runs/3191910918/jobs/5208774196) because the tests try to access to removed fields. For instance `testtuple.train.model`

tests/test_permissions.py
```diff
@@ -257,20 +259,22 @@ def test_permissions_model_process(
     )
 
     spec = factory.create_traintuple(
-        algo=algo_2, inputs=dataset_2.train_data_inputs + FLTaskInputGenerator.trains_to_train([traintuple_1.key])
+        algo=algo_2,
+        inputs=dataset_2.train_data_inputs
+        + FLTaskInputGenerator.trains_to_train([traintuple_1.key], worker=workers[1]),
```
sergebouchut2, line 264: Why adding worker arg as the method doesn't expect that and the input don't need this value?

sergebouchut2: OK I think I understand: the worker arg is for `create_traintuple` not `trains_to_train`

sergebouchut2 (approved):

PR #237, opened 2023-01-19
test: add test in case of add test task on org without training

ThibaultFy:


Owlfred:

tests/test_permissions.py
```diff
@@ -1,3 +1,5 @@
+from contextlib import contextmanager
```
guilhem-barthes, line 1: We cvan't use `from contextlib import nullcontext as does_not_raise` because you are using it in a context manager, right?

ThibaultFy: You're right, I saw this on a forum and was not sure about the `import  as does_not_raise` but didn't see it on the official doc on Python, let's use that instead !

PR #256, opened 2023-05-19
feat: use new substra functions

guilhem-barthes:
## Related issue

- https://github.com/Substra/substra-backend/pull/509
- https://github.com/Substra/substra/pull/361
- https://github.com/Substra/substrafl/pull/129

## Summary

Adapt tests to:
- remove usage of `.value` in task outputs
- use new functions to get input/output assets.

Fixes FL-652

## Test

Ci in `substra` companion PR

### DIFF
--- a/tests/test_execution.py
+++ b/tests/test_execution.py
@@ -42,11 +42,8 @@ def test_tasks_execution_on_same_organization(factory, network, client, default_
     assert traintask.error_type is None
     assert traintask.metadata == {"foo": "bar"}
     assert len(traintask.outputs) == 1
-    assert traintask.outputs[OutputIdentifiers.model].value is not None
-
-    if network.options.enable_model_download:
-        model = traintask.outputs[OutputIdentifiers.model].value
-        assert client.download_model(model.key) == b'{"value": 2.2}'
+    out_models = client.get_task_models(traintask.key)
+    assert len(out_models) == 1
 
     # check we can add twice the same traintask
     spec = get_traintask_spec()
@@ -121,10 +118,11 @@ def test_federated_learning_workflow(factory, client, default_datasets, workers)
         )
         traintask = client.add_task(spec)
         traintask = client.wait(traintask)
+        out_models = client.get_task_models(traintask.key)
         assert traintask.status == Status.done
         assert traintask.error_type is None
         assert len(traintask.outputs) == 1
-        assert traintask.outputs[OutputIdentifiers.model].value is not None
+        assert len(out_models) == 1
         assert traintask.tag == "foo"
         assert traintask.compute_plan_key  # check it is not None or ''
 
--- a/tests/test_execution.py
+++ b/tests/test_execution.py
@@ -330,8 +330,10 @@ if __name__ == '__main__':
     testtask = client.wait(testtask)
     assert testtask.status == Status.done
     assert testtask.error_type is None
-    assert testtask.outputs[identifier_1].value == 1
-    assert testtask.outputs[identifier_2].value == 2
+    output_1 = client.get_task_output_asset(testtask.key, identifier_1)
+    output_2 = client.get_task_output_asset(testtask.key, identifier_2)
+    assert output_1.asset == 1
+    assert output_2.asset == 2
 
 
 def test_testtask_with_same_output_identifer(factory, client):
--- a/tests/test_execution.py
+++ b/tests/test_execution.py
@@ -1,5 +1,8 @@
+from contextlib import nullcontext as does_not_raise
+
 import pytest
 import substra
+from substra.sdk.exceptions import TaskAssetNotFoundError
 from substra.sdk.models import Status
 from substra.sdk.schemas import AssetKind
 from substra.sdk.schemas import ComputeTaskOutputSpec
@@ -42,8 +45,8 @@ def test_tasks_execution_on_same_organization(factory, network, client, default_
     assert traintask.error_type is None
     assert traintask.metadata == {"foo": "bar"}
     assert len(traintask.outputs) == 1
-    output = client.get_task_output_asset(traintask.key, OutputIdentifiers.model)
-    assert output.asset is not None
+    with does_not_raise():
+        output = client.get_task_output_asset(traintask.key, OutputIdentifiers.model)
 
     if network.options.enable_model_download:
         model = output.asset
@@ -173,8 +176,8 @@ def test_tasks_execution_on_different_organizations(
     assert traintask.status == Status.done
     assert traintask.error_type is None
     assert len(traintask.outputs) == 1
-    output = client_1.get_task_output_asset(traintask.key, OutputIdentifiers.model)
-    assert output.asset is not None
+    with does_not_raise():
+        client_1.get_task_output_asset(traintask.key, OutputIdentifiers.model)
     assert traintask.worker == client_2.organization_id
 
     # add testtask; should execute on organization 1 (default_dataset_1 is located on organization 1)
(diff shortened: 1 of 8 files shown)


tests/test_execution.py
```diff
@@ -357,8 +366,11 @@ def test_composite_traintask_execution_failure(factory, client, default_dataset,
 
         assert composite_traintask.status == Status.failed
         assert composite_traintask.error_type == substra.sdk.models.TaskErrorType.execution
-        assert composite_traintask.outputs[OutputIdentifiers.local].value is None
-        assert composite_traintask.outputs[OutputIdentifiers.shared].value is None
+        with pytest.raises(ValueError):
+            client.get_task_output_asset(composite_traintask.key, OutputIdentifiers.local)
+
+        with pytest.raises(ValueError):
```
ThibaultFy, line 372: To change with the custom error

tests/test_execution.py
```diff
@@ -221,7 +226,8 @@ def test_function_build_failure(factory, network, default_dataset_1, worker):
 
         assert traintask.status == Status.failed
         assert traintask.error_type == substra.sdk.models.TaskErrorType.build
-        assert traintask.outputs[OutputIdentifiers.model].value is None
+        with pytest.raises(ValueError):
```
SdgJlbl, line 229: I feel like the `ValueError` is less readable than the `is None` was. 
Would it be possible to get the outputs and check that the model is None or that the key is not present? 
Same for below.

tests/test_execution_compute_plan.py
```diff
@@ -583,8 +585,8 @@ def test_compute_plan_aggregate_composite_traintasks(  # noqa: C901
     # Check that permissions were correctly set
     for task_id in [ct.task_id for ct in composite_traintask_specs]:
         task = clients[0].get_task(task_id)
-        trunk = task.outputs[OutputIdentifiers.shared].value
-        assert len(trunk.permissions.process.authorized_ids) == len(clients)
+        trunk = task.outputs[OutputIdentifiers.shared].permissions
```
ThibaultFy, line 588:
```suggestion
        trunk_permissions = task.outputs[OutputIdentifiers.shared].permissions
```

tests/test_execution_compute_plan.py
```diff
@@ -583,8 +585,8 @@ def test_compute_plan_aggregate_composite_traintasks(  # noqa: C901
...
+        assert len(trunk.process.authorized_ids) == len(clients)
```
ThibaultFy, line 589:
```suggestion
        assert len(trunk_permissions.process.authorized_ids) == len(clients)
```

tests/test_execution_compute_plan.py
```diff
@@ -561,7 +563,7 @@ def test_compute_plan_aggregate_composite_traintasks(  # noqa: C901
                 )
                 == 1
             )
-        print(task.inputs)
```
SdgJlbl, line 564: 🎉

tests/test_hybrid_mode.py
```diff
@@ -39,7 +39,8 @@ def test_execution_debug(client, hybrid_client, debug_factory, default_dataset):
     )
     traintask = hybrid_client.add_task(spec)
     assert traintask.status == models.Status.done
-    assert traintask.outputs[OutputIdentifiers.model].value is not None
+    output = hybrid_client.get_task_output_asset(traintask.key, OutputIdentifiers.model)
+    assert output.asset is not None
```
ThibaultFy, line 43: The real test should be:
```
from contextlib import nullcontext as does_not_raise
with does_not_raise():
   hybrid_client.get_task_output_asset(traintask.key, OutputIdentifiers.model)
```

tests/test_permissions.py
```diff
@@ -173,7 +173,8 @@ def test_permissions(permissions_1, permissions_2, expected_permissions, factory
     traintask = client_1.wait(traintask)
 
     # check the compute task executed on the correct worker
-    assert traintask.outputs[OutputIdentifiers.model].value is not None
+    output = client_1.get_task_output_asset(traintask.key, OutputIdentifiers.model)
+    assert output.asset is not None
```
ThibaultFy, line 177: same here with does_not_raise()

tests/workflows/mnist-fedavg/test_workflow.py
```diff
@@ -395,19 +395,26 @@ def test_mnist(factory, inputs, clients, cfg: Settings, workers: typing.List[str
     performances = client.get_performances(cp.key)
     perf_dict = {task_key: perf for task_key, perf in zip(performances.task_key, performances.performance)}
     testtasks = sorted(testtasks, key=lambda x: (x.rank, x.worker))
+
     for testtask in testtasks:
+        output = client.get_task_output_asset(testtask.key, OutputIdentifiers.performance)
```
ThibaultFy, line 400:
```suggestion
```

tests/workflows/mnist-fedavg/test_workflow.py
```diff
@@ -395,19 +395,26 @@ def test_mnist(factory, inputs, clients, cfg: Settings, workers: typing.List[str
...
         print(
             f"testtask({testtask.worker}) - rank {testtask.rank} "
             f"- round {testtask.metadata['round_idx']} "
-            f"perf: {perf_dict[testtask.key]}"
+            f"perf: {output.asset}"
```
ThibaultFy, line 404:
```suggestion
            f"perf: {perf_dict[testtask.key]}"
```

tests/workflows/mnist-fedavg/test_workflow.py
```diff
@@ -395,19 +395,26 @@ def test_mnist(factory, inputs, clients, cfg: Settings, workers: typing.List[str
...
         )
 
     nb_samples = (cfg.mnist_workflow.train_samples, cfg.mnist_workflow.test_samples)
 
     # performance should be deterministic a fixed number of samples:
     if nb_samples in _EXPECTED_RESULTS.keys():
         expected_perf = _EXPECTED_RESULTS[nb_samples]
-        assert all(perf_dict[testtask.key] == pytest.approx(perf) for (perf, testtask) in zip(expected_perf, testtasks))
+        assert all(
```
ThibaultFy, line 412: Revert changes ?

ThibaultFy (commented):
Thanks !! First review :)

tests/test_execution.py
guilhem-barthes: Thibault was suggesting creating errors (`AssetNotFoundError`) instead of `ValueError`. Would checking if it raises `AssetNotFoundError` clear for you?

SdgJlbl: Yes, I think that would be much clearer

tests/test_execution.py
```diff
@@ -173,8 +176,8 @@ def test_tasks_execution_on_different_organizations(
     assert traintask.status == Status.done
     assert traintask.error_type is None
     assert len(traintask.outputs) == 1
-    output = client_1.get_task_output_asset(traintask.key, OutputIdentifiers.model)
-    assert output.asset is not None
+    with does_not_raise():
```
SdgJlbl, line 179: do we need the `does_not_raise` here? 🤔

guilhem-barthes: We want to check that `get_task_output_asset` returns an object. This function will either raise an exception or return an object,  so checking if `is not None` seemed less clear to Thibault, as this `output.asset` will never be equals to `None`

SdgJlbl: then, we could just execute the function 🤷 
and add a comment that we just want to check that the function is not raising in this configuration

guilhem-barthes: Should I apply this change to all calls to `get_task_output_asset` (previously checking `assert output.asset is not None`), or you meant this one specifically?

guilhem-barthes: Should I apply this change to all calls to `get_task_output_asset` (previously checking `assert output.asset is not None`), or you meant this one specifically?

SdgJlbl: Using a context manager to check that a function does not raise seems pretty convoluted for me, so I would change all `with does_not_raise()` 
If we are checking that the correct Exception is raised, it makes sense to keep the context manager.

SdgJlbl (approved):

PR #259, opened 2023-06-22
fix: use `_token` instead of `token`

guilhem-barthes:
# Summary

Fix https://github.com/Substra/substra-tests/pull/257 where a `token` (attribute of `substratest.Client`) was still in use instead of `_token` (already available in `substra.Client`, now mother class of `substratest.Client`)


Owlfred:

ThibaultFy (approved):
thx :)

PR #364, opened 2024-07-11
chore: stop using substra-tools base images

SdgJlbl:


Makefile
```diff
@@ -14,7 +14,7 @@ test: test-remote test-local
 test-remote: test-remote-sdk test-remote-workflows
 
 test-remote-sdk: pyclean
-	pytest tests -rs -v --durations=0 -m "not workflows" -n $(PARALLELISM) --log-level=INFO
+	pytest tests -x -rs -v --durations=0 -m "not workflows" -n $(PARALLELISM) --log-level=INFO
```
SdgJlbl, line 17: TODO remove

tests/test_gpu.py
```diff
@@ -11,8 +11,8 @@ def test_gpu(factory, client, org_idx, default_datasets, workers):
     nvidia_drivers = "nvidiacuda11.8.0-base-ubuntu22.04"
 
     # Need the base image, the minimal image does not have pip
-    dockerfile = f"""
-FROM ghcr.io/substra/substra-tools:latest-{nvidia_drivers}-python3.7
+    dockerfile = """
+FROM python:3.12-slim
```
thbcmlowk, line 15: Isn't that slightly anticipated? 
```suggestion
FROM python:3.11-slim
```

substratest/settings.py
```diff
@@ -13,13 +13,11 @@
 _DEFAULT_NETWORK_CONFIGURATION_PATH = os.path.join(_CURRENT_DIR, "../", "values.yaml")
 _SUBSTRA_TESTS_CONFIG_FILEPATH = os.getenv("SUBSTRA_TESTS_CONFIG_FILEPATH", _DEFAULT_NETWORK_CONFIGURATION_PATH)
 
-_DEFAULT_SUBSTRA_TOOLS_TAG_LOCAL = (
-    f"latest-nvidiacuda11.8.0-base-ubuntu22.04-python{sys.version_info.major}.{sys.version_info.minor}"
-)
-_DEFAULT_SUBSTRA_TOOLS_TAG_REMOTE = "latest"
+_DEFAULT_DOCKER_TAG_LOCAL = f"{sys.version_info.major}.{sys.version_info.minor}-slim"
+_DEFAULT_DOCKER_TAG_REMOTE = "3.11-slim"
```
guilhem-barthes, line 17: Why do we need these two version, won't `_DEFAULT_DOCKER_TAG_LOCAL` return `3.11-slim` if started from the correct Python version? 

I can see that before, there was a difference in path between the two, but it looks like to me that we could use this opportunity to simplify the code

tests/test_gpu.py
SdgJlbl: Indeed, good catch

guilhem-barthes (approved):
🧹

PR #369, opened 2024-07-25
chore: bump Pytorch version to 2.2.1 in tests

SdgJlbl:
Fixes CI following Python 3.12 support.

### DIFF
--- a/changes/363.added
+++ b/changes/363.added
@@ -1,0 +1,1 @@
+Python 3.12 support.
--- a/changes/369.fixed
+++ b/changes/369.fixed
@@ -1,0 +1,1 @@
+Bump pytorch version to 2.2.1 in tests.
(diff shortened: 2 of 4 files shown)


thbcmlowk (approved):
Thanks for fixing this! How can we make sure we spot this earlier next time?

SdgJlbl:
> Thanks for fixing this! How can we make sure we spot this earlier next time?

I'm actually not sure 🤔 

The end-to-end tests were running fine, so I'm not sure... I guess that the matrices used are not exactly the same between the nightly and the manually triggered end-to-end tests, might be worth looking into it more deeply.

thbcmlowk:
> > Thanks for fixing this! How can we make sure we spot this earlier next time?
> 
> I'm actually not sure 🤔
> 
> The end-to-end tests were running fine, so I'm not sure... I guess that the matrices used are not exactly the same between the nightly and the manually triggered end-to-end tests, might be worth looking into it more deeply.

Yes indeed, I think the local tests are actually never triggered except in the nightly. Can you create a follow-up task so we can look at it later? Thank you!

Owlfred:

Owlfred:

PR #383, opened 2024-09-09
feat: add test for function build cancelation

guilhem-barthes:
Fixes FL-1659

## Companion PR

- https://github.com/Substra/substra-backend/pull/997

### DIFF
--- a/tests/test_docker_image_build.py
+++ b/tests/test_docker_image_build.py
@@ -110,15 +110,16 @@ def test_function_cancelled_cp(factory, cfg, client, worker, default_dataset):
 
     inputs = default_dataset.opener_input + default_dataset.train_data_sample_inputs
     cp = factory.create_compute_plan()
-    function = client.add_function(function_wait_spec)
-    cp.create_traintask(function=function, worker=worker, inputs=inputs)
-    function = client.add_function(function_test_spec)
-    cp.create_traintask(function=function, worker=worker, inputs=inputs)
+    function_wait = client.add_function(function_wait_spec)
+    cp.create_traintask(function=function_wait, worker=worker, inputs=inputs)
+    function_test = client.add_function(function_test_spec)
+    cp.create_traintask(function=function_test, worker=worker, inputs=inputs)
     client.add_compute_plan(cp)
 
     client.cancel_compute_plan(cp.key)
     cp = client.wait_compute_plan(cp.key, raise_on_failure=False, timeout=cfg.options.future_timeout)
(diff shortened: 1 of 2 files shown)


tests/test_docker_image_build.py
```diff
@@ -86,3 +86,39 @@ def test_function_build_order(factory, cfg, client, worker):
...
+    dockerfile_test = get_dockerfile(
+        docker_image, function_category, extra_instructions="ENV test=test_function_cancelled_cp_test"
+    )
+
+    function_wait_spec = factory.create_function(function_category, dockerfile=dockerfile_wait)
+    function_test_spec = factory.create_function(function_category, dockerfile=dockerfile_test)
+
+    inputs = default_dataset.opener_input + default_dataset.train_data_sample_inputs
+    cp = factory.create_compute_plan()
+    function = client.add_function(function_wait_spec)
```
SdgJlbl, line 113: WDYT of not calling both functions by the same variable name?

tests/test_docker_image_build.py
```diff
@@ -86,3 +86,39 @@ def test_function_build_order(factory, cfg, client, worker):
...
+    function = client.add_function(function_test_spec)
+    cp.create_traintask(function=function, worker=worker, inputs=inputs)
+    client.add_compute_plan(cp)
+
+    client.cancel_compute_plan(cp.key)
+    cp = client.wait_compute_plan(cp.key, raise_on_failure=False, timeout=cfg.options.future_timeout)
+    assert cp.status == "PLAN_STATUS_CANCELED"
+
+    function = client.wait_function(function.key, raise_on_failure=False, timeout=cfg.options.future_timeout)
+    assert function.status == "FUNCTION_STATUS_CANCELED"
```
SdgJlbl, line 124: We are not checking anything on the first function added 🤔

SdgJlbl (commented):
Thanks for your work, I think we are missing some status check on one of the two functions though.

tests/test_docker_image_build.py
guilhem-barthes: What do you think would make sense to check in this case?

tests/test_docker_image_build.py
guilhem-barthes: No strong opinion

tests/test_docker_image_build.py
SdgJlbl: That it's not failed or canceled, I guess?

SdgJlbl (approved):

